PR #51, opened 2023-01-05
Add --target-devops-center config variable

EstebanRomero84:
### What does this PR do?

- Adds new configuration variable called --target-devops-center.
- Makes --devops-center-username flag required.

[W-11940232](https://gus.lightning.force.com/a07EE000019NR6aYAG)

### Functionality Before

--devops-center-username was optional and no validations were enforced on it.

### Functionality After

We make sure there is a target devops center org. It could be set as a default org using the configuration variable or the user can provide one using the -c flag.


molson-sfdc (commented):
This is looking great.   So much has changed since I last worked in the CLI.  My only comment...is more code comments.

molson-sfdc (approved):
This looks good to me.  but maybe Nico/Juan should review as well.

npiccolo (approved):
Looks good to me Esteban !!
Sorry for the delay !!!

PR #57, opened 2023-01-10
feat: add logic to find the target pipeline stage id

EstebanRomero84:
### What does this PR do?
Adds logic to query the target org to get a pipelineStage record and compute the target pipelineStageId for a deployOnly promotion.

[W-11991578](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001BEvdjYAD/view)


src/common/selectors/pipelineStageSelector.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Connection } from '@salesforce/core';
+import { QueryResult } from 'jsforce';
+import { PipelineStage } from '../types';
+
+/**
+ *
+ * Returns an array of a pipeline stages filtering by project name.
+ */
+export async function selectPipelineStagesByProject(con: Connection, projectName: string): Promise<PipelineStage[]> {
+  const queryStr = `SELECT Id, sf_devops__Pipeline__r.sf_devops__Project__c, sf_devops__Branch__r.sf_devops__Name__c, (SELECT Id FROM sf_devops__Pipeline_Stages__r)
```
EstebanRomero84, line 17: We could also filter this query by branch name to get only the correct pipeline stage. But if the result is empty we can't know which one was wrong, the project name or the branch name. And it was a requirement to be very clear in the error message.

src/common/utils.ts
```diff
@@ -26,3 +27,36 @@ export function validateTestFlags(testLevel: Nullable<TestLevel>, tests: Nullabl
...
+    stages = await selectPipelineStagesByProject(targetOrg.getConnection(), projectName);
+  } catch (err) {
+    const error = err as ApiError;
+    if (error.errorCode === 'INVALID_TYPE') {
+      throw messages.createError('error.DevopsAppNotInstalled');
+    }
+    throw new SfError(getString(err, 'errorMessage', 'unknown'), getString(err, 'errorStatusCode', 'unknown'));
+  }
+  if (!stages.length) {
+    throw messages.createError('error.ProjectNotFound', [projectName]);
```
EstebanRomero84, line 52: It also could be the case that the project does exist but it hasn't pipelines associated to it. But I'm assuming that if the user is attempting a promotion then it has to have an active pipeline and pipeline stages.

src/common/selectors/pipelineStageSelector.ts
molson-sfdc: Can we validate the stage we found, to make sure it has the correct branch name?

src/common/types.ts
```diff
@@ -9,9 +9,33 @@ export type PromotePipelineResult = {
...
+export type Pipeline = {
+  sf_devops__Project__c: string;
+};
+
+export type PipelineStage = {
+  Id: string;
+  sf_devops__Branch__r: Branch;
+  sf_devops__Pipeline__r: Pipeline;
+  sf_devops__Pipeline_Stages__r?: PreviousPipelineStages;
+};
```
molson-sfdc, line 21: Future thought.  Can we share these type definitions with the Heroku app?  Might be nice one day.  Maybe a doce-types library?

src/common/utils.ts
molson-sfdc: We probably should enforce this.  Maybe verify sf_devops__Pipeline__r.sf_devops__Active__c is true?

EstebanRomero84: As discussed in Slack, we already have that validation on the PromoteService.

src/common/types.ts
EstebanRomero84: Good idea, I can make do a poc.

messages/commonErrors.md
```diff
@@ -0,0 +1,19 @@
+# error.NoTestsSpecified
+
+You must specify tests using the --tests flag if the --test-level flag is set to RunSpecifiedTests.
+
+# error.InvalidRunTests
+
+runTests can only be used with a testLevel of RunSpecifiedTests.
```
ekapner:
```suggestion
runTests can be used only with a testLevel of RunSpecifiedTests.
```

ekapner: @jshackell-sfdc - Esteban mentioned that he got the verbiage for this error from `sf deploy metadata` - I told him not to change it unless you feel like we should change it all places that use this error message. It's a minor geeky grammatical edit.

jshackell-sfdc: I agree with your edit, and I think Esteban should change it.  I just spent 5 minutes trying to find the corresponding error for `sf deploy metadata` and couldn't find it!   So I think it's fine to make yours better, and when I find the metadata ones I'll make them better too.

EstebanRomero84: Fixed. Thank you both.

ekapner (commented):
I approve the error message text change.

npiccolo (approved):
Looks good to me !!!
I really liked the abstractPromote class !!! will simplify lot of things !

PR #69, opened 2023-01-25
Better error message when the config variable does not correspond with an authenticated org 

EstebanRomero84:
### What does this PR do?
Update the validation for requiredDoceOrg flag so it prints a better error message when the username/alias stored in the -target-devops-center config variable does not correspond with an authenticated org.

[W-11940232](https://gus.lightning.force.com/a07EE000019NR6aYAG)

### Functionality Before

Error message says that no target org was provided.

### Functionality After

Error message says that the provided target org is not authenticated.

### DIFF
--- a/test/commands/deploy/pipeline.test.ts
+++ b/test/commands/deploy/pipeline.test.ts
@@ -8,7 +8,8 @@
 import { expect, test } from '@oclif/test';
 import { TestContext } from '@salesforce/core/lib/testSetup';
 import * as sinon from 'sinon';
-import { Org } from '@salesforce/core';
+import { ConfigAggregator, Org } from '@salesforce/core';
+import { ConfigVars } from '../../../src/configMeta';
 import { PromoteCommand } from '../../../src/common/abstractPromote';
 import { DeployPipelineCache } from '../../../src/common/deployPipelineCache';
 
@@ -30,6 +31,13 @@ describe('deploy pipeline', () => {
     sandbox = sinon.createSandbox();
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     sandbox.stub(Org, 'create' as any).returns(DOCE_ORG);
+    sandbox.stub(ConfigAggregator.prototype, 'getInfo').returns({
+      value: 'TARGET_DEVOPS_CENTER_ALIAS',
+      key: ConfigVars.TARGET_DEVOPS_CENTER,
+      isLocal: () => false,
+      isGlobal: () => true,
+      isEnvVar: () => false,
+    });
   });
 
   afterEach(() => {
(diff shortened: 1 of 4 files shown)


test/commands/deploy/pipeline.test.ts
```diff
@@ -117,6 +125,10 @@ describe('deploy pipeline', () => {
     test
       .stdout()
       .stderr()
+      .do(() => {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        executeCommandStub = sandbox.stub(PromoteCommand.prototype, 'executePromotion' as any);
```
JuanStenghele-sf, line 130: Why do we need this?

EstebanRomero84: I use it to mock the executePromotion() method from the base class. Since this is a unit test I want to verify the behaviour of _deploy pipeline_ command only and not the behaviour of the base class. I just assert that the method wass called.

JuanStenghele-sf: :nice:

test/common/flags.test.ts
```diff
@@ -134,4 +134,24 @@ describe('requiredDoceOrgFlag', () => {
       expect(err.message).to.include(`No authorization information found for ${invalidAlias}`);
     }
   });
+
+  it('fails when no alias is provided and the --target-devops-center config var is set but invalid', async () => {
+    const invalidTargetDevopsCenter = 'invalid-target-devops-center';
+    sandbox.stub(ConfigAggregator.prototype, 'getInfo').returns({
+      value: 'invalid-target-devops-center',
```
JuanStenghele-sf, line 141:
```suggestion
      value: invalidTargetDevopsCenter,
```

test/common/flags.test.ts
```diff
@@ -134,4 +134,24 @@ describe('requiredDoceOrgFlag', () => {
...
+      key: ConfigVars.TARGET_DEVOPS_CENTER,
+      isLocal: () => false,
+      isGlobal: () => true,
+      isEnvVar: () => false,
+    });
+    try {
+      await Parser.parse(['--requiredDoceOrg='], {
```
JuanStenghele-sf, line 148: I cannot find this flag. What is the difference from devops-center-username?

test/common/flags.test.ts
EstebanRomero84: In this case that it's a dummy hardcoded value for testing and variable naming conventions don't apply I like to use kebab casing for readability.

test/common/flags.test.ts
EstebanRomero84: Here I'm unit testing the callback function for the parser, it is not tied to any particular flag. So I opted to use a flag name that indicates what the callback does instead of using the flag name from our command.

test/common/flags.test.ts
JuanStenghele-sf: I understand. I thought we could also use the constant declared above here.

test/common/flags.test.ts
JuanStenghele-sf: Ok!

test/common/flags.test.ts
EstebanRomero84: Can't use the same constant because here I want to test the case where the given value doesn't match a valid authenticated org.

JuanStenghele-sf: Ok

JuanStenghele-sf (approved):
Nice job!

PR #76, opened 2023-02-02
feat: add logic to request the deployment

EstebanRomero84:
### What does this PR do?
Adds the logic to request the deployment.

[W-11912337](https://gus.lightning.force.com/a07EE000018phJuYAI/view)

### Functionality Before

The command does not make a request to initiate a deployment.

### Functionality After

The command makes a request to initiate a deployment.

### DIFF
--- a/src/commands/deploy/pipeline.ts
+++ b/src/commands/deploy/pipeline.ts
@@ -8,7 +8,8 @@
 import { Messages } from '@salesforce/core';
 import { SfCommand } from '@salesforce/sf-plugins-core';
 import { PromoteCommand } from '../../common/abstractPromote';
-import { PipelineStage, PromotePipelineResult } from '../../common';
+import { PipelineStage, PromoteOptions, PromotePipelineResult } from '../../common';
+import { APPROVED } from '../../common/constants';
 
 Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/plugin-devops-center', 'deploy.pipeline');
@@ -21,6 +22,7 @@ export default class DeployPipeline extends PromoteCommand<typeof SfCommand> {
   public static readonly description = messages.getMessage('description');
   public static readonly examples = messages.getMessages('examples');
   public static readonly state = 'beta';
+  private readonly isUndeployedOnly = true;
 
   public async run(): Promise<PromotePipelineResult> {
     return this.executePromotion();
@@ -30,17 +32,20 @@ export default class DeployPipeline extends PromoteCommand<typeof SfCommand> {
    * Computes the source stage Id for deployOnly promotions. If the given stage has a previous stage
    * then the source stage is the previous stage. If not the it means this is the
    * first stage of the pipeline and source stage Id = Approved.
+   *
+   * @returns: string. It is the source stage Id.
    */
-  protected computeSourceStageId(pipelineStage: PipelineStage): void {
-    this.sourceStageId = pipelineStage?.sf_devops__Pipeline_Stages__r
-      ? pipelineStage.sf_devops__Pipeline_Stages__r.records[0].Id
-      : 'Approved';
+  protected getSourceStageId(): string {
+    const targetStage: PipelineStage = this.getTargetStage();
+    return targetStage.sf_devops__Pipeline_Stages__r
+      ? targetStage.sf_devops__Pipeline_Stages__r.records[0].Id
+      : APPROVED;
   }
 
   /**
-   * Adds a promote option so it performs an undeployed only promotion
+   * Returns the promote option necessary to performs an undeployed only promotion
    */
-  protected addPromoteOptions(): void {
-    this.deployOptions = { ...this.deployOptions, undeployedOnly: true };
+  protected getPromoteOptions(): Partial<PromoteOptions> {
+    return { undeployedOnly: this.isUndeployedOnly };
   }
 }
--- a/src/commands/deploy/pipeline.ts
+++ b/src/commands/deploy/pipeline.ts
@@ -9,7 +9,6 @@ import { Messages } from '@salesforce/core';
 import { SfCommand } from '@salesforce/sf-plugins-core';
 import { PromoteCommand } from '../../common/abstractPromote';
 import { PipelineStage, PromotePipelineResult } from '../../common';
-import { DeployPipelineCache } from '../../common/deployPipelineCache';
 
 Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/plugin-devops-center', 'deploy.pipeline');
@@ -24,23 +23,24 @@ export default class DeployPipeline extends PromoteCommand<typeof SfCommand> {
   public static readonly state = 'beta';
 
   public async run(): Promise<PromotePipelineResult> {
-    // TODO Timeout case
-    if (this.flags.async) {
-      // TODO Get the aorId
-      const aorId = 'TODO';
-      await DeployPipelineCache.set(aorId, {});
-    }
     return this.executePromotion();
   }
 
   /**
-   * Computes the target stage Id for deployOnly promotions. If the given stage has a previous stage
-   * then the target stage is the previous stage. If not the it means this is the
-   * first stage of the pipeline and target stage Id = Approved.
+   * Computes the source stage Id for deployOnly promotions. If the given stage has a previous stage
+   * then the source stage is the previous stage. If not the it means this is the
+   * first stage of the pipeline and source stage Id = Approved.
    */
-  protected computeTargetStageId(pipelineStage: PipelineStage): void {
-    this.targetStageId = pipelineStage?.sf_devops__Pipeline_Stages__r
+  protected computeSourceStageId(pipelineStage: PipelineStage): void {
+    this.sourceStageId = pipelineStage?.sf_devops__Pipeline_Stages__r
       ? pipelineStage.sf_devops__Pipeline_Stages__r.records[0].Id
       : 'Approved';
+  }
+
+  /**
+   * Adds a promote option so it performs an undeployed only promotion
+   */
+  protected addPromoteOptions(): void {
+    this.deployOptions = { ...this.deployOptions, undeployedOnly: true };
   }
 }
--- a/src/common/abstractPromote.ts
+++ b/src/common/abstractPromote.ts
@@ -9,7 +9,13 @@ import { SfCommand } from '@salesforce/sf-plugins-core';
 import { Flags, Interfaces } from '@oclif/core';
 import { HttpRequest } from 'jsforce';
 import { DeployPipelineCache } from '../common/deployPipelineCache';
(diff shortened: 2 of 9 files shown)


src/commands/deploy/pipeline.ts
```diff
@@ -24,23 +23,24 @@ export default class DeployPipeline extends PromoteCommand<typeof SfCommand> {
...
+   * Computes the source stage Id for deployOnly promotions. If the given stage has a previous stage
+   * then the source stage is the previous stage. If not the it means this is the
+   * first stage of the pipeline and source stage Id = Approved.
    */
-  protected computeTargetStageId(pipelineStage: PipelineStage): void {
-    this.targetStageId = pipelineStage?.sf_devops__Pipeline_Stages__r
+  protected computeSourceStageId(pipelineStage: PipelineStage): void {
+    this.sourceStageId = pipelineStage?.sf_devops__Pipeline_Stages__r
       ? pipelineStage.sf_devops__Pipeline_Stages__r.records[0].Id
       : 'Approved';
```
molson-sfdc: Move to constants?

src/common/abstractPromote.ts
```diff
@@ -18,11 +20,20 @@ import {
...
 >;
 
+export type PromoteOptions = {
+  fullDeploy: boolean;
+  testLevel: string;
+  runTests: string;
+  undeployedOnly: boolean;
+  checkDeploy: boolean;
+  deploymentId: string;
+};
```
molson-sfdc: Should this be in types?

src/common/abstractPromote.ts
```diff
@@ -48,20 +61,58 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
+  private buildPromoteOptions(): Partial<PromoteOptions> {
+    // set base promote options
+    this.deployOptions = {
+      fullDeploy: this.flags['deploy-all'],
+      testLevel: this.flags['test-level'] ?? 'Default',
+      runTests: this.flags['tests'] ? this.flags['tests'].join(',') : undefined,
+    };
+    // ask concrete implementation to add more promote options if needed
+    this.addPromoteOptions();
+    return this.deployOptions;
```
molson-sfdc: Why is this stored on the instance and returned?  Do we need it after the promotion is requested?

src/commands/deploy/pipeline.ts
```diff
@@ -24,23 +23,24 @@ export default class DeployPipeline extends PromoteCommand<typeof SfCommand> {
...
   }
+
+  /**
+   * Adds a promote option so it performs an undeployed only promotion
+   */
+  protected addPromoteOptions(): void {
+    this.deployOptions = { ...this.deployOptions, undeployedOnly: true };
+  }
```
molson-sfdc, line 50: Minor comment about computeTargetStage and addPromoteOptions.  It is a bit safer to have the abstract class be the only one that deals with setting the member variables (and then they can be private).  In this case it might be a bit safer/cleaner if you had

computeTargetStageId(pipelineStage: PipelineStage): string

and

addPromoteOption(options: Partial\<PromoteOptions\>)

molson-sfdc (commented):
This is exciting to see!

I only have minor comments, and they are probably just from my ignorance of JS.  I don't need to see it again.  Nice work

src/commands/deploy/pipeline.ts
EstebanRomero84: Done.

src/common/abstractPromote.ts
EstebanRomero84: Yes. Moved.

src/commands/deploy/pipeline.ts
EstebanRomero84: Good to know, I changed it so the abstract class is the one that updates its members.

src/common/abstractPromote.ts
EstebanRomero84: Updated as part of below comment.

src/common/abstractPromote.ts
```diff
@@ -48,20 +59,66 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
 
   protected async executePromotion(): Promise<PromotePipelineResult> {
     validateTestFlags(this.flags['test-level'], this.flags.tests);
-    const doceOrg: Org = await Org.create({ aliasOrUsername: this.flags['devops-center-username']?.getUsername() });
-    const pipelineStage: PipelineStage = await fetchAndValidatePipelineStage(
+    const doceOrg: Org = this.flags['devops-center-username'] as Org;
```
JuanStenghele-sf, line 62: So, this is the same as calling `await Org.create({ aliasOrUsername: this.flags['devops-center-username']?.getUsername() });`?

src/common/abstractPromote.ts
```diff
@@ -48,20 +59,66 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
+
+  private async requestPromotion(targetOrg: Org): Promise<string> {
+    this.buildPromoteOptions();
+    const req: HttpRequest = {
+      method: 'POST',
+      url: `${REST_PROMOTE_BASE_URL as string}${
+        this.targetStage.sf_devops__Pipeline__r.sf_devops__Project__c
+      }/pipelineName/${this.sourceStageId}`,
+      body: JSON.stringify({
+        changeBundleName: this.flags['bundle-version-name'],
```
JuanStenghele-sf, line 94: What happens if it is not a bundling stage? Is this undefined?

src/common/abstractPromote.ts
```diff
@@ -48,20 +59,66 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
+        promoteOptions: this.deployOptions,
+      }),
+    };
+    return targetOrg.getConnection().request(req);
```
JuanStenghele-sf, line 98: Can this fail?

test/commands/deploy/pipeline.test.ts
```diff
@@ -136,4 +173,173 @@ describe('deploy pipeline', () => {
...
+            // now that we know the request was made
+            // we can get the call argument
+            // and validate its values
+            const requestArgument = requestMock.getCall(0).args[0] as HttpRequest;
+            expect(requestArgument.url).to.contain(REST_PROMOTE_BASE_URL);
+            const urlParams: string[] = requestArgument.url.split('/');
+            expect(urlParams[urlParams.length - 1]).to.equal(firstStageId);
+          }
+        );
+    });
```
JuanStenghele-sf, line 343: What about a test for checking that the test level default option is 'Default'?

JuanStenghele-sf (changes requested):
This looks really good, I left some questions.

src/common/abstractPromote.ts
EstebanRomero84: The end result is the same, we get the DOCe org. But with less processing because we don't create an new instance of an Org, we just us the one in the flag. And is safe to use `as Org` because we already validated it's an  Org instance with the flag parser.

src/common/abstractPromote.ts
EstebanRomero84: Correct. The type of this.flags['bundle-version-name'] is string | undefined.

src/common/abstractPromote.ts
EstebanRomero84: Yes, it can. If it fails we don't catch the exception,  we let it surface and then the cli will print the error message.

src/common/abstractPromote.ts
peternhale, line 62: @EstebanRomero84 I was looking at the implementation of `requiredDoceOrgFlag`. If you add property `required: true` you should be able to remove the `as Org` assertion.

Also, SfCommand `Flags` does export a `custom` method. I was curious why you are using OclifFlags version of that method.

```
export const requiredDoceOrgFlag = OclifFlags.custom({
  char: 'c',
  summary: messages.getMessage('flags.targetDoceOrg.summary'),
  parse: async (input: string | undefined) => getOrgOrThrow(input),
  default: async () => getOrgOrThrow(),
  defaultHelp: async () => (await getOrgOrThrow())?.getUsername(),
});
```

test/commands/deploy/pipeline.test.ts
EstebanRomero84: Good catch, I added it, thanks.

src/common/abstractPromote.ts
EstebanRomero84: Hi @peternhale - if we set it as required then the user will alway have to provide this flag right?  We want to give the user both possibilities: to specify an org using `-c` flag or defaulting to the `--target-devops-center` config variable.

EstebanRomero84: Regarding the custom flag, I based my implementation on this:
[https://github.com/salesforcecli/plugin-deploy-retrieve/blob/main/src/utils/flags.ts](https://github.com/salesforcecli/plugin-deploy-retrieve/blob/main/src/utils/flags.ts)

Also when I replace the OclifFlags flag by SfCommand flag I get an error: 

> Property 'custom' does not exist...

Am I missing something?

peternhale: Required does not mean the user has to provide a value, it just means that by the end of a the `await this.parse(<class>);` that there must be a value, either by use input or flag defaults.

peternhale: So this gives you an error?

```
export const requiredDoceOrgFlag = Flags.custom({
  char: 'c',
  summary: messages.getMessage('flags.targetDoceOrg.summary'),
  parse: async (input: string | undefined) => getOrgOrThrow(input),
  default: async () => getOrgOrThrow(),
  defaultHelp: async () => (await getOrgOrThrow())?.getUsername(),
});
```

EstebanRomero84: - I added the required attribute to requiredDoceOrgFlag and works as expected. Thanks.
- Yes, it gives me an error: 

> Property 'custom' does not exist on type '{ boolean: <T = boolean>(options?: Partial<BooleanFlag<T>> | undefined) => BooleanFlag<T>; directory: Definition<string, { exists?: boolean | undefined; }>; ... 10 more ...; optionalOrg: Definition

peternhale: @EstebanRomero84 check the import of Flags. Should be from `@salesforce/sf-plugins-core`

EstebanRomero84: As discussed in DM, I will update this flag once we upgrade to @salesforce/sf-plugin-core v2.

JuanStenghele-sf: Nice!

src/common/abstractPromote.ts
JuanStenghele-sf: Ok! I thought so

test/commands/deploy/pipeline.test.ts
JuanStenghele-sf: Nice

JuanStenghele-sf (approved):
Nice job, it looks really good!

PR #77, opened 2023-02-03
Stream

npiccolo:
### What does this PR do?

Adds the stream functionality 

@[W-11938846](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000019MizHYAS/view)@

### Functionality Before

None

### Functionality After

Streamer listening to AOR events.


messages/commonFlags.md
```diff
@@ -60,6 +60,14 @@ The command immediately returns the job ID and control of the terminal to you. T
 
 Run the command asynchronously.
 
+# promote.wait.description
+
+If the command continues to run after the wait period, the CLI returns control of the terminal window to you. To resume watching the deploy operation, run this command again. To check the status of the deploy operation, run "sf deploy pipeline report".
```
EstebanRomero84: This message is a bit different that the approved UIT. Did you check it with Emily? If so, please have the UIT updated.

npiccolo: Where is the approved UIT ?? This https://salesforce.quip.com/fxbMA99bbKIq ?

src/streamer/doceStream.ts
```diff
@@ -0,0 +1,65 @@
...
+      await asyncStatusClient.subscribe();
+    } catch (error) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      if (error.name === 'GenericTimeoutError') {
+        // We don't want log the error because we are expecting this timeout
+        // Maybe check if the timeout on the command input is the same as the timeout here ?
+        return;
+      } else {
+        // eslint-disable-next-line no-console
+        console.error(error);
```
EstebanRomero84: I think it should throw if catches an error different that the expected oen due to timeout.

src/streamer/processors/asyncOpStream.ts
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * This is the processor for the AOR CDC
+   *
+   * @param event The event JSON from the CDC
+   * @returns StatusResult
+   */
+  private startAsyncOpStreamProcessor(event: JsonMap): StatusResult {
+    const pl = ensureJsonMap(event.payload);
+
+    const c = ensureJsonMap(pl.ChangeEventHeader);
```
EstebanRomero84: This will be moved to the logger service once we have it right?

src/streamer/processors/asyncOpStream.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * It sends the correct config to the streamer
+   *
+   * @param asyncOpId the Id of the AOR we want to inspect
+   * @returns
+   */
+  public async startStreaming(asyncOpId: string): Promise<void | AnyJson> {
+    return this.startStream(
+      '/data/Async_Operation_Result__ChangeEvent',
```
EstebanRomero84: Move it to constants?

src/streamer/processors/asyncOpStream.ts
npiccolo: What do you mean by this ???
The console.log ???

test/common/flags.test.ts
```diff
@@ -103,35 +126,41 @@ describe('requiredDoceOrgFlag', () => {
...
-  it('fails when an invalid alias is provided and the --target-devops-center config var is not set', async () => {
-    const invalidAlias = 'invalidAlias';
-    sandbox.stub(ConfigAggregator.prototype, 'getInfo').returns({
-      value: null,
-      key: ConfigVars.TARGET_DEVOPS_CENTER,
-      isLocal: () => false,
-      isGlobal: () => true,
-      isEnvVar: () => false,
-    });
+describe('waitFlag', () => {
```
EstebanRomero84, line 160: Can we have a test to assert that --wait and --async flags are mutully exclusive?

test/streamer/asyncOpStream.test.ts
```diff
@@ -0,0 +1,186 @@
+/*
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { expect, test } from '@oclif/test';
```
EstebanRomero84, line 7: I think that here we should test the stream functionality only, in isolation. We shouldn't have to execute the deploy pipeline nor stub the fetchAndValidatePipelineStage() method.
Then in te test for deploy:pipeline we can stub the stream service and assert it was called.

src/streamer/processors/asyncOpStream.ts
EstebanRomero84: Yes. We are printing the message before checking it's the recordId we are expecting.

src/streamer/doceStream.ts
```diff
@@ -0,0 +1,65 @@
...
+      }
+    }
+  }
+
+  /**
+   * Returns the Id of the object we want to inspect
+   */
+  protected getIdtoInspect(): string {
+    return this.idToInspect;
```
molson-sfdc: You have a big assumption in this base class that all streaming implementations will be looking for a single object ID, and I do not think that is correct.  I'd like to see the reorganized a bit to avoid that assumption.

abstract class DOCeStreaming
   * This should deal with just the basics of subscribe/handshake/timeouts/etc.
 
abstract class SObjectStreaming extends DOCeStream
  * Takes the ID to look for in the constructor
  * Logic to determine if the ChangeEvent is for the given ID

 class AsycOpStreaming extends SObjectStreaming
  * Logic to determine if the AOR transitioned to a terminal state

src/common/abstractPromote.ts
```diff
@@ -56,6 +60,17 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
     this.computeTargetStageId(pipelineStage);
 
+    if (this.flags.async) {
+      // Empty waiting for logic
+      // We just return the aorId here
+      return { status: 'asyncStatus' };
+    }
+
+    const waitTimeout: Duration = this.flags.wait;
+    const streamer: AsyncOpStreaming = new AsyncOpStreaming(doceOrg, waitTimeout);
```
EstebanRomero84: Couldn't we just pass directly this.flags.wait as second parameter?

src/streamer/doceStream.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /**
+   * Sets the initial config for the stream
+   */
+  protected abstract startStreaming(idToInspect: string): Promise<void | AnyJson>;
```
molson-sfdc: Again, this assumes that all implementations of this will have a single ID to inspect.  That is probably not always going to be the case.  To avoid this, idToInspect should be passed into the constructor of the concrete class.  Then this interface can just be

startStreaming() : Promise<void | AnyJson>

src/streamer/processors/asyncOpStream.ts
molson-sfdc: Agreed.  And since it is constant it should probably be in the constructor

src/streamer/processors/asyncOpStream.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    const c = ensureJsonMap(pl.ChangeEventHeader);
+
+    // eslint-disable-next-line no-console
+    console.log(pl.sf_devops__Message__c);
+
+    if (
+      c.recordIds != null &&
+      c.recordIds[0] === this.getIdtoInspect() &&
+      (pl.sf_devops__Status__c === 'Completed' || pl.sf_devops__Status__c === 'Error')
```
molson-sfdc: Maybe "Status__c !== 'In Progress'" to be safer?  We have added "Ignored" status and will probably add "Canceled" soon.  Maybe AOR just needs a "Is_Completed__c" formula field so we can start to isolate this logic.  I'll add a story for that.

src/streamer/doceStream.ts
molson-sfdc: Yes, especially in the base class.  We do not know how the implementing class wants to deal with errors.

src/streamer/doceStream.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { Org, StreamingClient, StatusResult } from '@salesforce/core';
+import { AnyJson, JsonMap } from '@salesforce/ts-types';
+import { Duration } from '@salesforce/kit';
+
+export abstract class DOCeStreaming {
```
molson-sfdc: Just a thought, but do we want to put this behind an interface.  I know there are talks about maybe migrating this in the future to a different implementation.

molson-sfdc (changes requested):
The code looks good.  It is surprising how similar it is to the POC I showed in BA.

I would like to see a bit more separation of logic in the streaming class hierarchy. I think this will help keep things cleaner and easier to extend in the future.

src/streamer/sObjectStream.ts
```diff
@@ -0,0 +1,22 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { Org } from '@salesforce/core';
+import { Duration } from '@salesforce/kit';
+import DOCeStreaming from './doceStream';
+
+export default abstract class SObjectStreaming extends DOCeStreaming {
+  private idstoInspect: string[];
```
EstebanRomero84:
```suggestion
  private idsToInspect: string[];
```

src/commands/deploy/pipeline.ts
```diff
@@ -43,4 +43,18 @@ export default class DeployPipeline extends PromoteCommand<typeof SfCommand> {
...
       : 'Approved';
   }
+
+  /**
+   * Default function for catching commands errors.
+   *
+   * @param error
+   * @returns
+   */
+  protected catch(error: Error | SfError): Promise<SfCommand.Error> {
```
EstebanRomero84: Since this behivor is common to all implementations couldn't we moved this logic to the abstract class and then delegate the generation of the error message in the concrete classes?

EstebanRomero84 (commented):
Great work Nico! I just added a question and a small suggestion.

src/common/abstractPromote.ts
```diff
@@ -56,6 +59,16 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
     this.computeTargetStageId(pipelineStage);
 
+    if (this.flags.async) {
+      // Empty waiting for logic
+      // We just return the aorId here
+      return { status: 'asyncStatus' };
+    }
+
+    // harcoded value for the AOR until we get a real aorID
+    const streamer: AsyncOpStreaming = new AsyncOpStreaming(doceOrg, this.flags.wait, 'AORID');
```
molson-sfdc:
```suggestion
    const streamer: DoceMonitor = new AsyncOpStreaming(doceOrg, this.flags.wait, 'AORID');
```

molson-sfdc: We should hide all except the creation of this class behind the interface, so we can more easily swap the implementation later (if we desire)

src/streamer/doceStream.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { Org, StreamingClient, StatusResult } from '@salesforce/core';
+import { AnyJson, JsonMap } from '@salesforce/ts-types';
+import { Duration } from '@salesforce/kit';
+import Stream from './stream';
+export default abstract class DOCeStreaming implements Stream {
```
molson-sfdc:
```suggestion
/**
* Common base class for DOCe monitors that use the streaming api to......
**/
export default abstract class DOCeStreaming implements DoceMonitor {
```

src/streamer/doceStream.ts
```diff
@@ -0,0 +1,39 @@
...
+    this.wait = wait;
+  }
+
+  /**
+   * This class creates the client and starts listening to events.
+   *
+   * @param event The event we want to proccess (CDC)
+   * @param processor The especific processor we want to use for the event
+   */
+  public async startStream(event: string, processor: (message: JsonMap) => StatusResult): Promise<void | AnyJson> {
```
molson-sfdc:
```suggestion
  protected async startStream(event: string, processor: (message: JsonMap) => StatusResult): Promise<void | AnyJson> {
```

src/streamer/doceStream.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    const streamClient = await StreamingClient.create(options);
+    await streamClient.handshake();
+    await streamClient.subscribe();
+  }
+
+  /**
+   * Sets the initial config for the stream
+   */
+  protected abstract startStreaming(): Promise<void | AnyJson>;
```
molson-sfdc:
```suggestion
```

This is in the interface now...right?

src/streamer/doceStream.ts
molson-sfdc: Also, is event the correct parameter name?  Isn't it the channel?

src/streamer/processors/asyncOpStream.ts
```diff
@@ -0,0 +1,67 @@
...
+import SObjectStreaming from '../sObjectStream';
+import { ASYNC_OPERATION_CDC } from '../../common/constants';
+
+export default class AsyncOpStreaming extends SObjectStreaming {
+  public constructor(org: Org, wait: Duration, idToInspect: string) {
+    super(org, wait, new Array(idToInspect));
+  }
+
+  /**
+   * It sends the correct config to the streamer
```
molson-sfdc:
```suggestion
   * Connects to the AsyncOperationResults CDC channel and begins to minot for changes to the AOR we are monitoring.  If the AOR......
```

src/streamer/processors/asyncOpStream.ts
```diff
@@ -0,0 +1,67 @@
...
+   */
+  public async startStreaming(): Promise<void | AnyJson> {
+    return this.startStream(
```
molson-sfdc:
```suggestion
    return this.watchForSObject(
```

src/streamer/processors/asyncOpStream.ts
```diff
@@ -0,0 +1,67 @@
...
+      ASYNC_OPERATION_CDC,
```
molson-sfdc: This can be in the constructor

src/streamer/sObjectStream.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  protected constructor(org: Org, wait: Duration, idsToInspect: string[]) {
+    super(org, wait);
+    this.idstoInspect = idsToInspect;
+  }
+
+  protected isValidIdToInspect(idToInspect: string): boolean {
+    return this.idstoInspect.includes(idToInspect);
+  }
+}
```
molson-sfdc, line 49:
```suggestion
     protected watchForSObject(matchProcessor) {
       this.startStreaming(this.channelName, this.matchingPRocessor.bind();
     }
     protected matchingProcessor(event, matchProcessor) {
        //All the logic to determine if this event matches one of the IDS we are monitoring.
        if (weFoundAMatch) { matchPRocessor() }
     }
}
```

src/streamer/processors/asyncOpStream.ts
```diff
@@ -0,0 +1,67 @@
...
+    );
+  }
+
+  /**
+   * This is the processor for the AOR CDC
+   *
+   * @param event The event JSON from the CDC
+   * @returns StatusResult Completed: true => will end the stream.
+   */
+  protected startAsyncOpStreamProcessor(event: JsonMap): StatusResult {
```
molson-sfdc: The logic to determine if we matched an SObject should be in the super class (the SObject class).  Then this should only get called IF we match an ID and it should only contain the logic to see if we transitioned.  Maybe this just takes the jsonPayload as an argument?

src/streamer/stream.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright (c) 2023, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { StatusResult } from '@salesforce/core';
+import { AnyJson, JsonMap } from '@salesforce/ts-types';
+
+export default interface Stream {
```
molson-sfdc:
```suggestion
/**
* Common interface used to monitor the progress of something in the DevOps Center.
**/
export default interface DoceMonitor {
```

I think "Stream" is too dependent on the implementation.  It does not make as much sense if we implement a polling version of this...right?

src/streamer/stream.ts
```diff
@@ -0,0 +1,18 @@
...
+  /**
+   *
+   * @param event The CDC event we want to inspect
+   * @param processor The processor that will handle the event
+   */
+  startStream(event: string, processor: (message: JsonMap) => StatusResult): Promise<void | AnyJson>;
```
molson-sfdc:
```suggestion
  monitor(): Promise<void | AnyJson>;
```

Again, this should not talk about the implementation (or take implementation specific parameters).

molson-sfdc (changes requested):
Looking good.  I still think you have a bit too much logic in the AORStreaming class.  Also, the interface should not talk about streaming (as that is the implementation).  Most of my suggestions are just renaming things.

src/streamer/processors/asyncOpStream.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (changeEventHeader.recordIds != null && this.isValidIdToInspect(changeEventHeader.recordIds[0])) {
+      // in a future we want to test !Is_Completed__c
+
+      // Print the message from the payload if it exists
+      if (jsonPayload.sf_devops__Message__c) {
+        // eslint-disable-next-line no-console
+        console.log(jsonPayload.sf_devops__Message__c);
+      }
+
+      if (jsonPayload.sf_devops__Status__c !== 'In Progress') {
```
EstebanRomero84: Since the event contains only fields that have changed it is posible that sf_devops__Status__c is null, meaning that hasn't change and it is still "In Progress". So in this if statement we need to check that sf_devops__Status__c is not null.

messages/commonFlags.md
```diff
@@ -60,6 +60,14 @@ The command immediately returns the job ID and control of the terminal to you. T
 
 Run the command asynchronously.
 
+# promote.wait.description
+
+If the command continues to run after the wait period, the CLI returns control of the terminal window to you and returns the job ID. To check the status of the deploy operation, run "sf deploy pipeline report".
```
EstebanRomero84:
```suggestion
If the command continues to run after the wait period, the CLI returns control of the terminal window to you and returns the job ID. To check the status of the deploy operation, run "<%= config.bin %> <%= command.id %> report".
```

src/streamer/doceStream.ts
```diff
@@ -0,0 +1,40 @@
...
+  public constructor(org: Org, wait: Duration) {
+    this.doceOrg = org;
+    this.wait = wait;
+  }
+
+  /**
+   * This creates the streaming client and starts listening the desired streaming channel.
+   *
+   * @param channel The channel we want to listen (CDC)
+   * @param processor The especific processor we want to use for the event
```
EstebanRomero84:
```suggestion
   * @param processor The specific processor we want to use for the event
```

EstebanRomero84 (approved):
Looks great imo, I just requested minor changes.

src/streamer/doceStream.ts
```diff
@@ -0,0 +1,40 @@
...
+  protected async startStream(channel: string, processor: (message: JsonMap) => StatusResult): Promise<void | AnyJson> {
+    const options = new StreamingClient.DefaultOptions(this.doceOrg, channel, processor);
+    options.setSubscribeTimeout(this.wait);
+
+    const streamClient = await StreamingClient.create(options);
+    await streamClient.handshake();
+    await streamClient.subscribe();
+  }
+
+  public abstract monitor(): Promise<void | AnyJson>;
```
molson-sfdc, line 39: Q:  Is this needed, since it is defined in the interface that all concrete classes must implement?

src/streamer/sObjectStream.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * This method sets the processor we will need inside the matchingProcessor and will start the stream.
+   *
+   * @param matchProcessor This processor will be used in the matchingProcessor, it's provided by each implementation.
+   */
+  protected async watchForSObject(matchProcessor: (message: JsonMap) => StatusResult): Promise<void | AnyJson> {
+    this.matchProcessor = matchProcessor;
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+    return this.startStream(this.channelName, this.matchingProcessor.bind(this));
```
molson-sfdc: This feels a bit wrong setting matchProcessor on the instance.  It assumes it will not change.  I think that is valid for today, but maybe we can protect against it
```suggestion
    return this.startStream(this.channelName, (event: JsonMap) => {
      return this.matchingProcessor(event, matchProcessor);
    });
```

molson-sfdc (approved):
This looks good.  Thanks for all the changes.  I have one question and one minor suggestion.  I do not need to see it again.

src/streamer/doceStream.ts
npiccolo: Yes, else we get this error.

Class 'DOCeStreaming' incorrectly implements interface 'DoceMonitor'.
  Property 'monitor' is missing in type 'DOCeStreaming' but required in type 'DoceMonitor'.

molson-sfdc: Well that is stupid

src/streamer/sObjectStream.ts
npiccolo: Great !!!
And this is even easier to test, because the method we pass on the parameter will not be bounded I think !

npiccolo: Nope. is not easier to test the function we pass on parameter =/

PR #90, opened 2023-02-15
juan/W-11939334

JuanStenghele-sf:
### What does this PR do?

Creates the output service.

[W-11939334](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000019N74kYAC/view)

### Functionality Before

No output was shown for the deploy.

### Functionality After

New output service to print errors, status and deploy summaries.


src/common/outputService.ts
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  /**
+   * Prints the completed message if the deployment is completed
+   * or prints the error message if it failed
+   */
+  public static printDeploymentStatus(aor: AsyncOperationResult): void {
+    if (aor.sf_devops__Status__c === undefined || aor.sf_devops__Status__c === 'In Progress') {
+      console.log(aor.sf_devops__Message__c);
+    } else if (aor.sf_devops__Status__c === 'Completed') {
```
npiccolo: I think this message is already on the aor.message__c when it finishes correctly

src/common/outputService.ts
```diff
@@ -0,0 +1,192 @@
...
+      console.log(`Deploy failed./n${aor.sf_devops__Error_Details__c}`);
+    }
+  }
+
+  /**
+   * Prints an adHoc deploy summary
+   */
+  private static printAdHocDeploySummary(summary: DeploySummary): void {
+    console.log(
+      `DevOps Center pipeline stage ${summary.stageName} being updated with metadata associated with work ${
```
npiccolo: Maybe we want to have this kind of texts on a .md file ?

src/common/outputService.ts
```diff
@@ -0,0 +1,192 @@
...
+export class OutputService {
+  protected con: Connection;
+
+  public constructor(con: Connection) {
+    this.con = con;
+  }
+
+  public static printAsyncRun(aorId: string): void {
+    console.log(`Job ID: ${aorId}/n
+      Deploy has been queued./n
```
npiccolo: This is very specific for the deploy, I think that maybe we will have different messages for other operations with AORs

src/common/outputService/deployOutputService.ts
```diff
@@ -139,7 +139,7 @@ export class DeployOutputService extends OutputService {
         bundlesSummary.join('; '),
         summary.branchName,
         summary.orgUrl,
-      ])
+      ]) + '\n'
```
npiccolo: Did you check if you can adda an empty line on the .md file???

src/common/abstractResume.ts
```diff
@@ -69,14 +70,19 @@ export abstract class ResumeCommand<T extends typeof SfCommand> extends SfComman
...
 
     // it is resumable so we can start monitoring the operation
     this.log(`*** Resuming ${this.operationType} ***`);
     this.log(`Job ID: ${bold(asyncJobId)}`);
-    const streamer: DoceMonitor = getAsyncOperationStreamer(doceOrg, this.flags.wait, asyncJobId);
+    const streamer: DoceMonitor = getAsyncOperationStreamer(
+      doceOrg,
+      this.flags.wait,
+      asyncJobId,
+      OutputServiceFactory.forDeployment(doceOrg.getConnection())
```
npiccolo: This abstract resume class will be more generic than they deploy I think

src/common/outputService/deployOutputService.ts
JuanStenghele-sf: Yes, but this is the only working option that I found. I will try to look for alternatives.

src/common/abstractResume.ts
JuanStenghele-sf: Why don't we make printAorStatus static so as to avoid having to instantiate a class?

src/common/selectors/endpointSelector.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Connection } from '@salesforce/core';
+
+type OrgUrlQueryResult = {
+  Endpoint: string;
+};
+
+/**
+ * Given a named credential it returns the org url (endpoint) associated with it
+ */
+export async function selectOrgUrl(con: Connection, namedCredential: string): Promise<string> {
```
EstebanRomero84:
```suggestion
export async function selectNamedCredentialByName(con: Connection, name: string): Promise<NamedCredential> {
```

src/common/selectors/endpointSelector.ts
```diff
@@ -0,0 +1,27 @@
...
+  const queryStr = `
+    SELECT Endpoint
+    FROM NamedCredential
+    WHERE DeveloperName = '${namedCredential}'
+    AND NamespacePrefix = 'sf_devops'
+    LIMIT 1`;
+  const resp: QueryResult<OrgUrlQueryResult> = await con.query(queryStr);
+  return resp.records[0].Endpoint;
```
EstebanRomero84: I think this method should return a NamedCredential record. Then the caller decides which field to use.

src/common/abstractPromote.ts
```diff
@@ -77,18 +79,25 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
+
       return {
         jobId: asyncOperationId,
         status: AsyncOperationStatus.InProgress,
       };
+    } else {
+      await outputService.printProgressSummary(asyncOperationId, this.flags['branch-name']);
+
+      const doceMonitor: DoceMonitor = new AsyncOpStreaming(doceOrg, this.flags.wait, asyncOperationId, outputService);
+      await doceMonitor.monitor();
```
molson-sfdc: This falls through and on line 100 we monitor again.  Is that intentional?

src/streamer/processors/asyncOpStream.ts
```diff
@@ -33,18 +41,19 @@ export default class AsyncOpStreaming extends SObjectStreaming {
...
-    if (jsonPayload.sf_devops__Message__c) {
-      // eslint-disable-next-line no-console
-      console.log(jsonPayload.sf_devops__Message__c);
-    }
+    // We build an aor given the payload
+    const asyncOpResult: AsyncOperationResult = {
+      Id: jsonPayload.Id as string,
+      sf_devops__Error_Details__c: jsonPayload.sf_devops__Error_Details__c as string,
+      sf_devops__Status__c: jsonPayload.sf_devops__Status__c as AsyncOperationStatus,
+      sf_devops__Message__c: jsonPayload.sf_devops__Message__c as string,
```
molson-sfdc: Can't you just cast jsonPayload to a AsyncOperationResult?   They have the same members.

src/common/abstractPromote.ts
```diff
@@ -77,18 +79,25 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
     this.sourceStageId = this.getSourceStageId();
     const asyncOperationId: string = await this.requestPromotion(doceOrg);
 
-    // TODO: move this to logger service
-    this.log(`Job ID: ${asyncOperationId}`);
-    await DeployPipelineCache.set(asyncOperationId, {});
+    const outputService: OutputService = OutputServiceFactory.forDeployment(doceOrg.getConnection());
```
molson-sfdc: I would use constructor injection for this (and the resume command).  It will make it easier to test.

src/common/abstractResume.ts
molson-sfdc: :no:  statics are so difficult to test.

src/common/outputService/outputService.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { green, red } from 'chalk';
+import { AsyncOperationResult } from '../types';
+
+/**
+ * Service class to print the output
+ *
+ * @author JuanStenghele-sf
+ */
+export abstract class OutputService {
```
molson-sfdc: I think this class hierarchy should be broken up a bit more.

abstract class OutputService ==> Only things that are used by all commands (maybe nothing now)
abstract class AorOutputService extends OutputService ==> interfaces that deal with an AOR
interface ResumeOutputService.  Generic things needed by resume logic.  Like printProgressSummary
class ResumeCommandOutputService extends AorOutputService implements ResumeOutputService 
interface DeployOutputService.  Generic things about deployment
class DeployCommandOutputService extends AorOutputService implements DeployOutputService, ResumeOutPutService

src/common/outputService/outputService.ts
```diff
@@ -0,0 +1,48 @@
...
+    } else if (aor.sf_devops__Status__c === 'Error' && aor.sf_devops__Error_Details__c) {
+      console.log(red(`${aor.sf_devops__Message__c}\nError details: ${aor.sf_devops__Error_Details__c}`));
+    }
+  }
+
+  /**
+   * Prints the aor id
+   */
+  public printAorId(aorId: string): void {
+    console.log(`Job ID: ${aorId}\n`);
```
molson-sfdc: aorId is constant and can be passed in the constructor or set once (in AorOutputService)

src/common/outputService/outputService.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * Prints a summary for an async run
+   */
+  public abstract printAsyncRunInfo(aorId: string): void;
+
+  /**
+   * Prints the progress summary
+   */
+  public abstract printProgressSummary(aorId: string, branch: string): Promise<void>;
```
molson-sfdc: branch is also constant and only relevant in the DeployOutoutService, so setting it should be part of that interface.

src/common/outputService/deployOutputService.ts
```diff
@@ -0,0 +1,261 @@
...
+  stageName: string;
+  orgUrl: string;
+};
+
+/**
+ * Service class to print the deploy output
+ *
+ * @author JuanStenghele-sf
+ */
+export class DeployOutputService extends OutputService {
```
molson-sfdc: See my comment below on the OutputService class hierarchy.  Besides that, I think this needs to be broken up.  This class should just deal with outputting information when asked.  A new class hierarchy something like

interface DeploySummaryProvider {
  DeploySummary createSummary();
}
class AdHocPromoteSummaryProvider implements SummaryProvider {}
class SoupPromoteSummaryProvider implements SummaryProvider {}
class VersionedPromoteSummaryProvider implements SummaryProvider {}

src/common/outputService/deployOutputService.ts
```diff
@@ -0,0 +1,261 @@
...
+  protected con: Connection;
+
+  public constructor(con: Connection) {
```
molson-sfdc: After the above change, this constructor should be

public constructor(provider: DeploySummaryProvider)

molson-sfdc (changes requested):
Let's look at breaking up some of this logic into different class hierarchies.  I think that will improve reusability and make testing easier.

src/common/outputService/deployOutputService.ts
```diff
@@ -0,0 +1,261 @@
...
+   * Builds a DeploySummary for an adHoc deploy
+   */
+  private async processAdHocDeploy(branchName: string, workItemsPromotes: WorkItemPromote[]): Promise<DeploySummary> {
+    const stageName: string = workItemsPromotes[0].sf_devops__Pipeline_Stage__r.Name;
+
+    const workItems: string[] = workItemsPromotes.map((workItem) => workItem.sf_devops__Work_Item__r.Name);
+
+    const namedCredential: string =
+      workItemsPromotes[0].sf_devops__Pipeline_Stage__r.sf_devops__Environment__r.sf_devops__Named_Credential__c;
+    const orgUrl: string = await selectOrgUrl(this.con, namedCredential);
```
molson-sfdc: As you move to a SummaryProvider, keep in mind _a lot_ of this information is static over the life of the promotion.  We should be able to compute it once.

src/common/abstractPromote.ts
JuanStenghele-sf: It was a merge problem, it is now fixed.

src/streamer/processors/asyncOpStream.ts
JuanStenghele-sf: :ok:

src/common/outputService/outputService.ts
JuanStenghele-sf: printProgressSummary  is not just for the ResumeOutputService, we also use it for the deploy itself. What would you put in ResumeOutputService?

src/common/outputService/deployOutputService.ts
JuanStenghele-sf: How do you know which type of provider to send? Because now we check that inside the output service class (we detect this after the first query)

molson-sfdc: Ahh, good point.....  We won't know that at least until we load the source pipeline stage.

Either pass a DeploySummerFactory into the constructor, or create the DeploySummary when this class is aware of the pipeline stage (or when we get the first AOR).

src/common/outputService/outputService.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Service interface to print the output
+ *
+ * @author JuanStenghele-sf
+ */
+export interface OutputService {
+  /**
+   * Prints a summary of the operation being done
+   */
+  printOpSummary(): Promise<void>;
```
EstebanRomero84: Two questions regarding this method:
- Could it have a more generic name, like print()? In the future we may have commands that need to print stuff that is not necessarily a summary. 
- If its main responsibility is to print a message, why it returns a promise? Does it need to asynchronous?

JuanStenghele-sf: - The idea of the method is to print a summary of the operation done, so that's why it has summary in its name. The idea would be in the future with more output services add more methods (if it is not a summary).
- It is async because we gather the information needed for the summary (For example, queries). However, the queries are made in other classes that are called from here so this must be async because its dependencies are async.

src/common/abstractPromote.ts
```diff
@@ -77,18 +85,26 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
-    this.log(`Job ID: ${asyncOperationId}`);
-    await DeployPipelineCache.set(asyncOperationId, {});
+    this.outputService.setAorId(asyncOperationId);
+
+    this.outputService.printAorId();
+    await this.outputService.printOpSummary();
 
     if (this.flags.async) {
-      // TODO display async message
+      await DeployPipelineCache.set(asyncOperationId, {});
```
npiccolo: I think this was moved outside the if statement because we always need to set the cache.
Do you remember this @EstebanRomero84 ?

npiccolo: I remember why now !!
If we want to resume an operation that timeout, we need to have that aorid on the cache to resume it using the most recent

EstebanRomero84: Yes, you are right.

src/common/abstractResume.ts
```diff
@@ -18,6 +17,8 @@ import {
 import { jobId, requiredDoceOrgFlag, useMostRecent, wait } from '../common/flags';
 import DoceMonitor from '../streamer/doceMonitor';
 import { DeployPipelineCache } from './deployPipelineCache';
+import { OutputServiceFactory } from './outputService/outputServiceFactory';
+import { ResumeCommandOutputService } from './outputService/resumeCommandOutputService';
```
npiccolo: Maybe we can add an Index.ts for output, so we just need to import all from ./outputService only ???

src/common/abstractResume.ts
```diff
@@ -59,24 +64,28 @@ export abstract class ResumeCommand<T extends typeof SfCommand> extends SfComman
...
     }
 
     // it is resumable so we can start monitoring the operation
-    this.log(`*** Resuming ${this.operationType} ***`);
-    this.log(`Job ID: ${bold(asyncJobId)}`);
-    const streamer: DoceMonitor = getAsyncOperationStreamer(doceOrg, this.flags.wait, asyncJobId);
+    this.outputService.setAorId(asyncJobId);
+
+    await this.outputService.printOpSummary();
+    this.outputService.printAorId();
```
npiccolo, line 85: We don’t need to print the aorid before ?

src/common/types.ts
```diff
@@ -43,6 +45,30 @@ export interface ApiError extends Error {
...
+};
+
+export type WorkItemPromote = {
+  sf_devops__Pipeline_Stage__r: PipelineStage;
+  sf_devops__Work_Item__r: WorkItem;
+};
+
+export type Environment = {
+  Id: string;
+  sf_devops__Named_Credential__c: string;
```
npiccolo, line 59: This is not a NamedCredential type ??

src/common/abstractPromote.ts
JuanStenghele-sf: Ok! I will move this outside the if.

src/common/abstractResume.ts
JuanStenghele-sf: You mean creating an index.ts file in the outputService directory and adding there just the imports for everything?

src/common/abstractResume.ts
JuanStenghele-sf: The output was like this before the PR. I don't know exactly.

src/common/types.ts
JuanStenghele-sf: In our environment object__c the named credential is just a string, thats why here we only store this (we don't know the other fields of the named credential).

src/common/outputService/deployCommandOutputService.ts
```diff
@@ -9,8 +9,8 @@
 
 import { SfCommand } from '@salesforce/sf-plugins-core';
 import { Flags } from '../abstractPromote';
-import { AbstractPromoteOutputService } from './promoteOutputService';
 import { DeploySummaryBuilder } from './deploySummaryBuilder';
+import { AbstractPromoteOutputService } from './promoteOutputService';
```
npiccolo, line 13: LOL

src/common/abstractResume.ts
npiccolo: Before the PR we didn’t have a printSummary =P

src/common/types.ts
npiccolo: okok

src/common/abstractResume.ts
JuanStenghele-sf: Yes, but we had the console.logs with that, I moved them inside the resume output service

npiccolo (approved):
Looks good to me.
Some minor changes that you already did =P
I would wait for Esteban's review though

src/common/abstractResume.ts
npiccolo: Ahhhh yes !!! My bad it's the resuming !!! I thought was the abstract promote class !!!
Sorry

src/common/outputService/outputService.ts
EstebanRomero84: Thanks for the answer.

- Got it.
- IMO the return type in the interface shouldn't be a promise, that is an implementation detail that the concrete classes should define. Otherwise all implementations are required to return a promise even if it's not needed, for example ResumeCommandOutputService.printOpSummary(). I think is safe to make the interface and abstract classes to return void and the concrete classes like DeploySummaryOutputService and AbstractPromoteOutputService to return Promise<void>. Open for discussion.

JuanStenghele-sf: Ok! I fixed this in the last commit.

EstebanRomero84 (approved):
Great work Juan!

src/common/selectors/deployProgressSummarySelector.ts
```diff
@@ -0,0 +1,35 @@
...
+  aorId: string
+): Promise<DeploySummaryQueryResult> {
+  const queryStr = `SELECT Id, sf_devops__Operation__c,
+                    (SELECT sf_devops__Pipeline_Stage__r.Name, sf_devops__Pipeline_Stage__r.sf_devops__Environment__r.sf_devops__Named_Credential__c, sf_devops__Work_Item__r.Name FROM sf_devops__Work_Item_Promotes__r),
+                    (SELECT sf_devops__Change_Bundle__r.Id, sf_devops__Change_Bundle__r.sf_devops__Version_Name__c, sf_devops__Environment__r.Id FROM sf_devops__Change_Bundle_Installs__r),
+                    (SELECT Name FROM sf_devops__Work_Items__r)
+                    FROM sf_devops__Async_Operation_Result__c 
+                    WHERE Id = '${aorId}'`;
+  const resp: QueryResult<DeploySummaryQueryResult> = await con.query(queryStr);
+  return resp.records[0];
```
molson-sfdc, line 34: I created [W-12624312](https://gus.lightning.force.com/a07EE00001MEHm1YAH) for us to come back in and harden all of the selectors.  They should not assume records has results.

molson-sfdc (approved):
These changes look good.  Thanks for reworking them

src/common/selectors/deployProgressSummarySelector.ts
JuanStenghele-sf: Ok!

PR #91, opened 2023-02-16
New command deploy pipeline resume

EstebanRomero84:
### What does this PR do?
Adds new command `deploy pipeline resume`. It is used to resume an in progress deploy operation.
`deploy pipeline report` will be added in another PR.

[W-11912492](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000018pwRAYAY/view)


src/common/utils.ts
```diff
@@ -67,3 +71,21 @@ export async function fetchAndValidatePipelineStage(
...
+    const error = err as Error;
+    if (error.name === 'SingleRecordQuery_NoRecords') {
+      throw messages.createError('error.InvalidAorId', [aorId]);
+    }
+    throw err;
+  }
+  return aor;
+}
+
+export function getAsyncOperationStreamer(org: Org, waitTime: Duration, idToInspect: string): AsyncOpStreaming {
```
EstebanRomero84, line 89: I added this method so I can stub AsyncOpStreaming.

src/common/deployPipelineCache.ts
```diff
@@ -77,12 +77,13 @@ export class DeployPipelineCache extends TTLConfig<TTLConfig.Options, AsyncOpera
...
   /**
-   * Returns the latest data cached
+   * Returns the latest data cached or throws if can't find any.
    */
-  public resolveLatest(): string {
+  public getLatestKeyOrThrow(): string {
     const aorId = this.getLatestKey();
-    if (!aorId) throw cacheMessages.createError('error.NoRecentAorId');
-
+    if (!aorId || aorId === 'undefined') {
```
EstebanRomero84: This is weird, the getLatestKey() method returns the string literal 'undefined' when it doesn't find any key.

messages/commonFlags.md
```diff
@@ -72,6 +72,28 @@ Number of minutes to wait for command to complete and display results.
 
 Username or alias of the target org.
 
+# flags.job-id.summary
+
+Job ID of the deploy operation you want to resume.
```
npiccolo: Maybe remove some specific things like deploy here in the commonFlags ?

messages/commonFlags.md
```diff
@@ -72,6 +72,28 @@ Number of minutes to wait for command to complete and display results.
...
+
+# flags.job-id.description
+
+These commands return a job ID if they time out or you specified the --async flag:
+
+- sf deploy pipeline
```
npiccolo, line 83: Maybe use _<%= config.bin %> <%= command.id %>_ and we can reuse it for promotes in a future ?

src/common/deployPipelineCache.ts
peternhale: @EstebanRomero84 
Using the branch for this PR I added this test case to `test/common/deployPipelineCache.test.ts` and the function does not return 'undefined' for a key miss, it returns undefined.

```  it('return undefined when trying to get latest key in a new cache', async () => {
    const cache = await DeployPipelineCache.create();
    const entry = cache.getLatestKey();
    expect(entry).to.not.be.ok;
  });
```

src/common/flags.ts
```diff
@@ -93,6 +93,20 @@ export const requiredDoceOrgFlag = OclifFlags.custom({
   required: true,
 });
 
+export const jobId = Flags.salesforceId({
+  char: 'i',
```
peternhale, line 97: You can add the Id prefix and the parse will validate it.

messages/commonFlags.md
EstebanRomero84: Good idea, I'll update this.

messages/commonFlags.md
EstebanRomero84: In this case I don't think is necessary because here we want to list all commands that can be resumed. SO in the future when we have another commands we can add the to the list, not to repleace it.

src/common/deployPipelineCache.ts
EstebanRomero84: You are right, I'll fix it.
Thanks for the review.

src/common/flags.ts
EstebanRomero84: This flag is the ID for our **AsyncOperationResult** object. Since it is a custom object, will it have the same prefix in all orgs?

peternhale: Yes, it will.

kylewalke: Standard Objects will have consistent prefixes in their ID's but Custom Objects (which our package installs) will not.

src/common/utils.ts
```diff
@@ -89,3 +89,36 @@ export async function fetchAsyncOperationResult(con: Connection, aorId: string):
...
+
+/**
+ *
+ * Helper to compare to SF Ids independently if they are 15 or 18 characters long.
+ *
+ * @param firstId
+ * @param secondId
+ * @returns true is the Ids match
+ */
+export function matchesSfId(firstId: string, secondId: string): boolean {
```
npiccolo, line 122: <3

npiccolo (approved):
Looks good to me !!!
thank you very much Esteban !!!

PR #97, opened 2023-02-22
Retry to deploy when 409 thrown

kylewalke:
Retry up to 50 times by default when a
user attempts to run the deploy command
and gets back a 409 error.

### What does this PR do?
Retry up to 50 times by default when a
user attempts to run the deploy command
and gets back a 409 error.

https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001L01NhYAJ/view

### Functionality Before
No retry, just thrown error. 

### Functionality After
Retries on 409 errors up to 50 times before moving on.


src/common/abstractPromote.ts
```diff
@@ -79,6 +85,25 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
+    try {
+      const asyncOperationId: string = await this.requestPromotion(doceOrg);
+      return asyncOperationId;
+    } catch (error) {
+      const err = error as HttpResponse;
+      // if we get anything besdies a 409 or run out of retries we throw an error
+      if (err['errorCode'] !== 'ERROR_HTTP_409' || numRetries < 1) {
+        return JSON.stringify(err); // we can't throw the error from this promise because then it doesn't surface to the user
+      }
+      // this.log('Number of Retries Remaining: ' + numRetries.toString()); // todo: do we want to remove this for users since we won't tell them how many times we are trying/failing ??
```
kylewalke: these two lines will be removed

src/common/abstractPromote.ts
```diff
@@ -95,6 +120,19 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
         promoteOptions: this.deployOptions,
       }),
     };
+    /*
```
kylewalke: this will also be deleted but is handy for local testing

src/common/abstractPromote.ts
```diff
@@ -75,7 +76,11 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
       this.flags['branch-name']
     );
     this.sourceStageId = this.getSourceStageId();
-    const asyncOperationId: string = await this.requestPromotion(doceOrg);
+    const asyncOperationId: string = await this.retryEnabledRequestPromotion(doceOrg, this.numRetries409);
+    this.spinner.stop();
+    if (asyncOperationId?.includes('errorCode')) {
+      throw new Error(JSON.stringify(asyncOperationId)); // Is this the type of error we want to throw here?
```
kylewalke: If there's a specific type of other error we want to construct and throw, then here's the place to do it. otherwise I'll delete my comment here

molson-sfdc: IIRC the responses should contain approved UIT.  right?

src/common/abstractPromote.ts
```diff
@@ -101,6 +106,23 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
 
+  protected async retryEnabledRequestPromotion(doceOrg: Org, numRetries: number): Promise<string> {
+    try {
+      const asyncOperationId: string = await this.requestPromotion(doceOrg);
+      return asyncOperationId;
+    } catch (error) {
+      const err = error as HttpResponse;
+      // if we get anything besdies a 409 or run out of retries we throw an error
+      if (err['errorCode'] !== 'ERROR_HTTP_409' || numRetries < 1) {
+        return JSON.stringify(err); // we can't throw the error from this promise because then it doesn't surface to the user
```
molson-sfdc: Wouldn't it be better to change the return type of this method to `Promise<string|HttpResponse>` ?  Then the caller would have full access to the error (and could rethrow it if needed)

src/common/abstractPromote.ts
```diff
@@ -101,6 +106,23 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
+      }
+      if (numRetries === this.numRetries409)
+        // we only start the spinner the first time we pass through here
+        this.spinner.start('Sync of VCS Events In-Progress');
```
molson-sfdc: I am a little concerned with this logic.  a). the logic to start/stop the spinner is separated, so it is difficult to follow.  b) it is dependent on the member variable `numRetries409` not changing.

Just thinking out loud.  Can we maybe break the functionality up a bit?  

retryEnabledRequestPromotion(org) {
   try {
      return this.tryOnce(org);
  } catch (EventNotDoneException e) {
    this.startSpinner();
    try {
      return this.tryWithLoop(org. this.num409s-1);
    finally {
      this.stopSpinner();
    } catch (error) {
      return error as HttpResponse;
    }
}

tryWithLoop(org, numRetries) {
    try {
      return this.tryOnce(org);
    } catch (EventNotDoneException e) {
      return this.tryLoop(org, numRetries - 1);
    }
}

tryOnce(org) {
  try {
    return await this.requestPromotion(org);
   } cach (error) {
     if (error == RETRY) throw new EventNotDoneException()
     return error;
  }
}


There is also an interesting read here on recursive Promises.  Which might help with some of the error handling issues you saw:  https://stackoverflow.com/questions/29020722/recursive-promise-in-javascript

molson-sfdc (changes requested):
See my comments below.  Open for discussion

src/common/abstractPromote.ts
kylewalke: Yeah, this was something I talked over with Esteban just now and I changed how I'm doing this so it's handled better.

src/common/abstractPromote.ts
kylewalke: Sure! I can rewrite for this. Also figured a better way to pass through errors that I didn't know worked before.

kylewalke: I broke it up into 2 different methods similar to what you have here. I think it looks much better than before so hopefully it's easier to read this way.

src/common/abstractPromote.ts
```diff
@@ -101,6 +102,38 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
     };
   }
 
+  protected async requestPromotionFlow(doceOrg: Org, numRetries: number): Promise<string> {
```
molson-sfdc: Method docs please (both methods)

src/common/abstractPromote.ts
```diff
@@ -101,6 +102,38 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
+    let asyncOperationId: string;
+    try {
+      asyncOperationId = await this.requestPromotion(doceOrg);
+    } catch (error) {
+      const err = error as HttpResponse;
```
molson-sfdc: Here(and on line 127) will this always be ab HttpResponse?  What if this.requestPromotion throws an NPE?

src/common/abstractPromote.ts
```diff
@@ -101,6 +102,38 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
+      // if we get anything besides a 409 we rethrow the error
+      if (err['errorCode'] !== HTTP_CONFLICT_CODE) {
+        throw error;
+      }
+      this.spinner.start('Sync of VCS Events In-Progress');
+      asyncOperationId = await this.retryRequestPromotion(doceOrg, numRetries);
```
molson-sfdc: Wrap this in a try/finally and stop the spinner in the finally.  Then you do not need to also do it on line 130.  It will also catch _all_ of the cases.

src/common/abstractPromote.ts
```diff
@@ -101,6 +102,38 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
+      this.spinner.stop();
+    }
+    return asyncOperationId;
+  }
+
+  protected async retryRequestPromotion(doceOrg: Org, numRetries: number): Promise<string> {
+    try {
+      const asyncOperationId: string = await this.requestPromotion(doceOrg);
```
molson-sfdc:
```suggestion
      return await this.requestPromotion(doceOrg);
```

Minor nit, no sense taking up extra heap space, especially in a recursive method.

test/commands/deploy/pipeline.test.ts
```diff
@@ -698,6 +702,78 @@ describe('deploy pipeline', () => {
...
+          sf_devops__Pipeline__r: {
+            sf_devops__Project__c: 'mockProjectId',
+          },
+          sf_devops__Pipeline_Stages__r: undefined,
+        };
+        fetchAndValidatePipelineStageStub = sandbox
+          .stub(Utils, 'fetchAndValidatePipelineStage')
+          .resolves(pipelineStageMock);
+      });
+      test
```
molson-sfdc, line 820: Can we start a best practice of documenting the test cases.  Sometimes they can be difficult to grok.  It does not need to be much.  In this case you have comments in the test setup, which helps, but a summary would be even better when scanning.

//Test the case where we get 49 CONFLICT errors and on the 50th we get a different error

test/commands/deploy/pipeline.test.ts
```diff
@@ -698,6 +702,78 @@ describe('deploy pipeline', () => {
...
+          // on the 4th try we want to complete the VCS event processing and return an AOR Id to the user
+          requestMock.onCall(4).resolves('mock-aor-id');
+        })
+        .command(['deploy:pipeline', '-p=testProject', '-b=testBranch'])
+        .it('Succeeds the request after processing some VCS Events', (ctx) => {
+          // make sure we made the callout the right number of times
+          expect(requestMock.callCount).to.equal(5);
+          // make sure that we show the returned aor id to the user
+          expect(ctx.stdout).to.contain('mock-aor-id');
+        });
```
molson-sfdc, line 893: I _think_ you are missing the case where the first call returns an AORid.

molson-sfdc (changes requested):
Much easier to read and follow.  Thanks.  A few minor comments, but overall looking good.

src/common/abstractPromote.ts
EstebanRomero84: @molson-sfdc I'll be working in this PR.
I find these two methods to be somewhat repetitive. Here is my suggestion to have the retry logic in one method.
```suggestion
protected async requestPromotionWithRetry(doceOrg: Org, numRetries: number): Promise<string> {
    try {
      return await this.requestPromotion(doceOrg);
    } catch (error) {
      if (error instanceof Error && error['errorCode'] === HTTP_CONFLICT_CODE && numRetries >= 1 ) {
        this.spinner.start('Sync of VCS Events In-Progress');
        return await this.requestPromotionWithRetry(doceOrg, numRetries - 1);
      }
      throw error;
    } finally {
      // we need to stop the spinner in case it was started in a retry attempt.
      this.spinner.stop();
    }
  }
```

src/common/abstractPromote.ts
```diff
@@ -101,6 +102,38 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
+    let asyncOperationId: string;
+    try {
+      asyncOperationId = await this.requestPromotion(doceOrg);
+    } catch (error) {
+      const err = error as HttpResponse;
+      // if we get anything besides a 409 we rethrow the error
+      if (err['errorCode'] !== HTTP_CONFLICT_CODE) {
+        throw error;
+      }
+      this.spinner.start('Synchronization of source control system events in progress');
```
EstebanRomero84: UIT approved by Emily.

src/common/abstractPromote.ts
molson-sfdc: Won't this repeatedly call `startSpinner` ?

EstebanRomero84: Yes, it will.
We could use `this.spinner.status` to check if it is already spinning and start it if it's not.

src/common/abstractPromote.ts
```diff
@@ -101,6 +102,52 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
 
+  /**
+   *
+   * Sends an Http request to the target org to initiate a promotion.
+   *
+   * @param doceOrg target org.
+   * @returns Aor Id of promote operation.
+   */
+  protected async requestPromotionFlow(doceOrg: Org): Promise<string> {
+    try {
```
molson-sfdc, line 130:
```suggestion
    let spinnerStarted = false;
    try {
```

src/common/abstractPromote.ts
```diff
@@ -101,6 +102,52 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
+      return await this.requestPromotion(doceOrg);
+    } catch (error) {
+      const err = error as ApiError;
+      // if we get a 409 error then call the retry flow
+      if (err.errorCode === HTTP_CONFLICT_CODE) {
+        this.spinner.start('Synchronization of source control system events in progress');
+        return await this.retryRequestPromotion(doceOrg, this.numRetries409);
```
molson-sfdc, line 138:
```suggestion
        spinnerStarted = true;
        return await this.retryRequestPromotion(doceOrg, this.numRetries409);
```

src/common/abstractPromote.ts
```diff
@@ -101,6 +102,52 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
+      }
+      throw error;
+    } finally {
+      this.spinner.stop();
```
molson-sfdc:
```suggestion
      if (spinnerStarted) this.spinner.stop();
```

If we get an exception (not 409) in the first call, this will call this.spinner.stop() without it being started...right?  I am not sure if that is bad or not, but you can guard against it.

molson-sfdc (approved):
These changes look good. Thanks.  Just one question about stopSpinner maybe getting called w/o start.  I don't need to see this again.

PR #103, opened 2023-03-02
Deploy pipeline report command

EstebanRomero84:
### What does this PR do?

Adds new command _sf deploy pipeline report_.

[W-11912492](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000018pwRAYAY/view)

### DIFF
--- a/src/common/outputService/reportOutputService.ts
+++ b/src/common/outputService/reportOutputService.ts
@@ -8,9 +8,9 @@
 import { SfCommand } from '@salesforce/sf-plugins-core';
 import { CliUx } from '@oclif/core';
 import { Flags } from '../abstractReportOnPromote';
-import { tableHeader } from '../outputUtils';
-import { DeploymentResult } from '../types';
 import { sObjectToArrayOfKeyValue } from '../utils';
+import { DeploymentResult } from '../types';
+import { tableHeader } from './outputUtils';
 import { AbstractDeploymentResultOutputService, DeploymentResultOutputService } from './deploymentResultOutputService';
 import { OutputFlags } from './outputService';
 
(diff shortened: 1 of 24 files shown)


src/common/outputService/reportOutputService.ts
```diff
@@ -0,0 +1,67 @@
...
+        x.key = x.key === 'Name' ? 'CreatedByName' : x.key;
+        return x;
+      })
+      // sort by key.
+      .sort((a, b) => (a.key < b.key ? -1 : 1));
+
+    const options = { title: tableHeader(`${this.operationName} Info`) };
+    const columns = { key: { header: 'Key' }, value: { header: 'Value' } };
+
+    CliUx.ux.table(formattedDeploymentResult, columns, options);
```
molson-sfdc, line 65: Will we want this logic in the `sf deploy` and `sf resume` commands as well?

molson-sfdc (commented):
This looks good.  Just one question (mostly based on my ignorance)

src/common/outputService/reportOutputService.ts
EstebanRomero84: In those commands we will print a table with the deployed components. But the title and columns will be different.

molson-sfdc: OK, Cool.  My only minor comment then, is do you need `printDeployResults` as a separate method?  Your abstract command is just calling printOpSummary(), which I think is fine.

molson-sfdc (approved):
These look great.  One minor question, but I do not need to see it again

src/common/outputService/reportOutputService.ts
EstebanRomero84: My thinking was that maybe in the future we want report commands to display more info, for example deployed components. So `printOpSummary()` would call `printDeploymentResult() `and `printDeployedComponents()`

npiccolo: I'm working on this printDeployedComponents().
It will just be a method calling this CliUx.ux.table with the correct parameters (columns and title defined for deployedcomponents table).
We only need to pass the rows we want to display to this method.

PR #110, opened 2023-03-08
Npiccolo/verbosity

npiccolo:
### What does this PR do?

Adds verbose/concise flags

@[W-12498420](https://gus.lightning.force.com/a07EE00001KlWoyYAF)@
@[W-12624312](https://gus.lightning.force.com/a07EE00001MEHm1YAH)@

### Functionality Before

No flags

### Functionality After

Flags and functionality added.
Concise: Will display a short summary
Verbose: Will display a table with the deployed components at the end of the run


src/common/selectors/deployComponentsSelector.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function selectDeployComponentsByAsyncOpId(
+  con: Connection,
+  asyncOpId: string
+): Promise<DeployComponent[]> {
+  const queryStr = `SELECT sf_devops__Source_Component__c, sf_devops__Operation__c, sf_devops__File_Path__c
+                    FROM sf_devops__Deploy_Component__c  
+                    WHERE sf_devops__Deployment_Result__r.sf_devops__Status__c = '${asyncOpId}'`;
+  const resp: QueryResult<DeployComponent> = await con.query(queryStr);
+
+  resp.records.forEach((record) => {
```
EstebanRomero84: I think we should move this logic from the selector to the caller in case this method is called from other places in the future.

src/common/table.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { CliUx } from '@oclif/core';
+
+export function displayTable(
+  rows: any[],
+  title: string,
+  columns: CliUx.Table.table.Columns<any>,
+  options?: CliUx.Table.table.Options
+): void {
+  CliUx.ux.log();
```
EstebanRomero84: What does this line do?

src/common/types.ts
```diff
@@ -113,3 +113,11 @@ export enum AsyncOperationStatus {
 export type NamedCredential = {
   Endpoint: string;
 };
+
+export type DeployComponent = {
```
EstebanRomero84, line 117: Could be do something like this so DeployComponent type is reusable?
```suggestion
export type DeployComponent = {
  sf_devops__Source_Component__c: string;
  sf_devops__Operation__c: string;
  sf_devops__File_Path__c: string;
};

export type FormattedDeployComponent = DeployComponent & {
  Type?: string;
  Name?: string;
};
```

src/common/table.ts
npiccolo: Adds an empty line, just to separate the final result from command line

src/common/types.ts
npiccolo: Is reusable DeployComponent, Type and Name are optional in this case.

src/common/selectors/deployComponentsSelector.ts
npiccolo: It can be used from other places, I thought this was the best option because in most cases we want the type and name separately. If we move this logic outside potentially we are going to have that logic called from different places.

EstebanRomero84: Could we have a utility method to perform the formatting? So if you need the original records you call the selector and if you need them formatted you call the utility method.

src/common/base/abstractPromote.ts
```diff
@@ -107,6 +108,9 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
       this.outputService
     );
     await doceMonitor.monitor();
+    if (this.outputService.getStatus() === AsyncOperationStatus.Completed && this.flags.verbose) {
+      await this.outputService.displayEndResults();
```
molson-sfdc, line 116: I think we should let the OuputService decide what to display.  It knows the verbosity level right?  

if (getStatus() === Completed)
  this.outputService.displayEndResults();

src/common/outputService/outputService.ts
```diff
@@ -20,6 +24,16 @@ export interface OutputService {
...
+
+  /**
+   * Displays a table
+   */
+  displayTable(
+    rows: any[],
+    title: string,
+    columns: CliUx.Table.table.Columns<any>,
+    options?: CliUx.Table.table.Options
+  ): void;
```
molson-sfdc: Will this ever be called external to the OutputService?  It does not feel like it.  I would remove it from the interface, but keep it in the abstract class.

src/common/outputService/promoteOutputService.ts
```diff
@@ -62,4 +70,12 @@ export abstract class AbstractPromoteOutputService
       await deploySummaryOutputService.printOpSummary();
     }
   }
+
+  /**
+   * This method will print a table of the deployed components for the current AOR
+   */
+  public async displayEndResults(): Promise<void> {
+    const components: DeployComponent[] = await getFormattedDeployComponentsByAyncOpId(this.con, this.aorId);
```
molson-sfdc:
```suggestion
     if (this.flags.verbose) {
         const components: DeployComponent[] = await getFormattedDeployComponentsByAyncOpId(this.con, this.aorId);
           ....
```

molson-sfdc (commented):
This looks pretty good.  Only a few minor comments.  I don't need to see it again.

EstebanRomero84 (approved):
Looks fine imo.

package.json
```diff
@@ -21,7 +21,7 @@
     "@salesforce/prettier-config": "^0.0.2",
     "@salesforce/ts-sinon": "1.4.6",
     "@swc/cli": "^0.1.62",
-    "@swc/core": "^1.3.39",
+    "@swc/core": "1.3.39",
```
npiccolo, line 24: We removed the caret just to take the exact version of this library.

Peter: When regenerating yarn.lock the dep @swc/core@^1.3.39 picks up v 1.3.40. That version is somehow breaking snapshot:compare and generate. Pin @swc/core@1.3.39 at exact version.

EstebanRomero84 (approved):
Great work.

PR #121, opened 2023-03-17
feat: query hardening

npiccolo:
### What does this PR do?

Adds some verifications for queries.

@[W-12624312](https://gus.lightning.force.com/a07EE00001MEHm1YAH)@

### Functionality Before

No verifications for queries.

### Functionality After

Verification for connection, queryString, noReultsFound and queryFailed.

### DIFF
--- a/src/common/base/abstractReportOnPromote.ts
+++ b/src/common/base/abstractReportOnPromote.ts
@@ -72,7 +72,7 @@ export abstract class ReportOnPromoteCommand<T extends typeof SfCommand> extends
       };
     } catch (err) {
       const error = err as Error;
-      if (error.name === 'SingleRecordQuery_NoRecords') {
+      if (error.name === 'No-results-foundError') {
         throw messages.createError('error.InvalidAorId', [asyncJobId]);
       }
       throw err;
--- a/src/common/base/abstractReportOnPromote.ts
+++ b/src/common/base/abstractReportOnPromote.ts
@@ -61,7 +61,7 @@ export abstract class ReportOnPromoteCommand<T extends typeof SfCommand> extends
         this.operationName
       );
       displayer.printOpSummary();
-    } else if (operationResult === null) {
+    } else if (!operationResult) {
       throw messages.createError('error.InvalidAorId', [asyncJobId]);
     }
 
--- a/src/common/base/abstractReportOnPromote.ts
+++ b/src/common/base/abstractReportOnPromote.ts
@@ -49,33 +49,27 @@ export abstract class ReportOnPromoteCommand<T extends typeof SfCommand> extends
       : (this.flags['job-id'] as string);
     const org = this.flags['devops-center-username'] as Org;
 
-    try {
-      // query the deployment result related to the promote operation.
-      const operationResult: DeploymentResult = await selectOneDeploymentResultByAsyncJobId(
-        org.getConnection(),
-        asyncJobId
+    // query the deployment result related to the promote operation.
+    const operationResult: DeploymentResult | null = await selectOneDeploymentResultByAsyncJobId(
+      org.getConnection(),
+      asyncJobId
+    );
+    if (operationResult && !this.jsonEnabled()) {
+      const displayer: OutputService = new OutputServiceFactory().forPromotionReport(
+        this.flags,
+        operationResult,
+        this.operationName
       );
-      if (!this.jsonEnabled()) {
-        const displayer: OutputService = new OutputServiceFactory().forPromotionReport(
-          this.flags,
-          operationResult,
-          this.operationName
-        );
-        displayer.printOpSummary();
-      }
-
-      return {
-        jobId: asyncJobId,
-        status: operationResult.sf_devops__Status__r.sf_devops__Status__c,
-        message: operationResult.sf_devops__Status__r.sf_devops__Message__c,
-        errorDetails: operationResult.sf_devops__Status__r.sf_devops__Error_Details__c,
-      };
-    } catch (err) {
-      const error = err as Error;
-      if (error.name === 'No-results-foundError') {
-        throw messages.createError('error.InvalidAorId', [asyncJobId]);
-      }
-      throw err;
+      displayer.printOpSummary();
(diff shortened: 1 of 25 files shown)


src/common/selectors/selectorUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+    throw messages.createError('error.connection-required');
+  }
+  if (typeof queryStr !== 'string' || queryStr.trim().length === 0) {
+    throw messages.createError('error.query-string-required');
+  }
+
+  // Run the query and verify for results
+  try {
+    const result: QueryResult<T> = await con.query(queryStr);
+    if (result.totalSize > 0 && result.records) {
```
molson-sfdc: In our Heroku app we also check 'result.done'.  Is that needed here?

Also, do we have any any queries (or will we) that could return an empty list?

src/common/selectors/selectorUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+      return result;
+    }
+  } catch (error) {
+    throw messages.createError('error.query-failed');
+  }
+
+  // If we arrive here, it means the query has no results.
+  throw messages.createError('error.no-results-found');
```
molson-sfdc, line 39: I think QueryResult can contain error information, do we want to log that any where?

molson-sfdc (commented):
Overall looks good.  Some questions.

src/common/selectors/selectorUtils.ts
EstebanRomero84: Yes, fetchAsyncOperationResult() when we filter by the Id provided by the user in resume.

npiccolo: If we get here a result.done = false, do we want to do a querymore() ???

molson-sfdc: I still think this should support an optional parameter to allow it to return a null result (when totalSize === 0).   Then we can harden the logic in abstractReportOnPromote

src/common/base/abstractReportOnPromote.ts
```diff
@@ -72,7 +72,7 @@ export abstract class ReportOnPromoteCommand<T extends typeof SfCommand> extends
       };
     } catch (err) {
       const error = err as Error;
-      if (error.name === 'SingleRecordQuery_NoRecords') {
+      if (error.name === 'No-results-foundError') {
```
molson-sfdc: This seems brittle.  What if another query in this try block fails?  See my comment below.  I think selectOneDeploymentResultByAsyncJobId should be allowed to return null.

molson-sfdc (changes requested):

src/common/base/abstractReportOnPromote.ts
npiccolo: Okok !!
So we need that boolean to **allowNullResults** on **runSafeQuery**.
By default we can set it to true, and we can pass it as false for AbstractReportOnPromote for example.
It seems ok for you ??

src/common/base/abstractReportOnPromote.ts
```diff
@@ -49,33 +49,27 @@ export abstract class ReportOnPromoteCommand<T extends typeof SfCommand> extends
...
-    try {
-      // query the deployment result related to the promote operation.
-      const operationResult: DeploymentResult = await selectOneDeploymentResultByAsyncJobId(
-        org.getConnection(),
-        asyncJobId
+    // query the deployment result related to the promote operation.
+    const operationResult: DeploymentResult | null = await selectOneDeploymentResultByAsyncJobId(
+      org.getConnection(),
+      asyncJobId
+    );
```
molson-sfdc, line 56:
```suggestion
    );
    if (!operationResult) {
      throw...
    }
```

Minor, just be very explicit about it.  I think it will be easier to follow the logic below.

src/common/base/abstractReportOnPromote.ts
```diff
@@ -49,33 +49,27 @@ export abstract class ReportOnPromoteCommand<T extends typeof SfCommand> extends
...
+    if (operationResult && !this.jsonEnabled()) {
+      const displayer: OutputService = new OutputServiceFactory().forPromotionReport(
+        this.flags,
+        operationResult,
```
molson-sfdc, line 60:
```suggestion
        operationResult!,
```

Then I think you need this to correct the type of operationResult

molson-sfdc (approved):
Just a minor not.  The rest looks good.  I do not need to see it again.

src/common/base/abstractReportOnPromote.ts
peternhale: Typescript should understand that operationResult is not undefined, given the guard testing operationResult. Using `!` is an anti-pattern that should be avoided.

peternhale (commented):
Just a suggestion

src/common/base/abstractReportOnPromote.ts
npiccolo: In this case operationResult it exists or it's null at this point, because we previously tested the existence inside the query method and we send back the correct result.

This is the return from the query method:
`  return resp.totalSize > 0 ? resp.records[0] : null;`

src/common/selectors/selectorUtils.ts
```diff
@@ -0,0 +1,40 @@
...
+  queryStr: string,
+  allowNullResults = false
+): Promise<QueryResult<T>> {
+  // Parameters verification
+  if (!con) {
+    throw messages.createError('error.connection-required');
+  }
+  if (typeof queryStr !== 'string' || queryStr.trim().length === 0) {
+    throw messages.createError('error.query-string-required');
+  }
```
peternhale, line 25: @npiccolo Typescript knows these are not undefined and of the correct type. Why the extra checks?

npiccolo: I did the connection check because we can have potentially things like this:
`this.flags['devops-center-username'] as Org).getConnection()`, this can't fail at some point and pass something wrong to the method ???
And the queryString I tested it because I have already tested the other parameter

PR #122, opened 2023-03-17
Albatross prerelease

EstebanRomero84:
### What does this PR do?

Updates GH actions to execute a prerelease.

### DIFF
--- a/.github/workflows/onPushToMain.yml
+++ b/.github/workflows/onPushToMain.yml
@@ -3,12 +3,19 @@ name: version, tag and github release
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      # point at specific branches, or a naming convention via wildcard
+      - prerelease
+    tags-ignore:
+      - '*'
 
 jobs:
   release:
     uses: salesforcecli/github-workflows/.github/workflows/githubRelease.yml@main
     secrets: inherit
+    with:
+      prerelease: ${{ github.ref_name != 'main' }}
 
   # most repos won't use this
   # depends on previous job to avoid git collisions, not for any functionality reason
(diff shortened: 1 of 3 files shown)


.github/workflows/onPushToMain.yml
```diff
@@ -3,12 +3,19 @@ name: version, tag and github release
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      # point at specific branches, or a naming convention via wildcard
+      - prerelease
```
iowillhoit, line 9: Are you only wanting this to run on the `prerelease` branch or branches that start with `prerelease`? If it's the latter, this would need to be `prerelease/*`

.github/workflows/onRelease.yml
```diff
@@ -2,20 +2,31 @@ name: publish
...
 on:
   release:
-    types: [released]
-  # support manual release in case something goes wrong and needs to be repeated or tested
+    types: [published]
   workflow_dispatch:
     inputs:
       tag:
-        description: tag that needs to publish
+        description: prerelease-albatross
```
iowillhoit: This description does not need to be specific to prereleases. You could manually run this workflow with a tag (prerelease or not) to trigger a publish

iowillhoit:
Added a few comments, but other than that I think this is correct. I am happy to troubleshoot it you have any issues.

.github/workflows/onPushToMain.yml
EstebanRomero84: Just prerelease branch.

.github/workflows/onRelease.yml
EstebanRomero84: Got it, I updated it.

iowillhoit (approved):

molson-sfdc (approved):

PR #131, opened 2023-03-23
fix: noResultsFound on project

npiccolo:
### What does this PR do?

Fix a little issue with error message when project not found

None

### Functionality Before

Bad message

### Functionality After

Good message


EstebanRomero84 (approved):
Looks good. Thank you Nico.

PR #139, opened 2023-03-29
fix: prevent defaultHelp method to throw when there is no target-devops-center value set

EstebanRomero84:
### What does this PR do?
Change the callback function on requiredDoceOrgFlag.defaultHelp. The new callback does not throw if it can't find a target Devops Center org.

[W-12738813](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001NfaZ7YAJ/view)

### Functionality Before

Running any command without having a value in target-devops-center config variable fails with error  _"You must specify the DevOps Center org username by indicating the -c flag on the command line or by setting the target-devops-center configuration variable"_ even when the user does provide -c flag.

### Functionality After

Commands correctly detect the provided -c flag.

### DIFF
--- a/src/common/flags/flags.ts
+++ b/src/common/flags/flags.ts
@@ -47,7 +47,7 @@ export const requiredDoceOrgFlag = OclifFlags.custom({
   summary: messages.getMessage('flags.targetDoceOrg.summary'),
   parse: async (input: string | undefined) => getOrgOrThrow(input),
   default: async () => getOrgOrThrow(),
-  defaultHelp: async () => (await getOrgOrThrow())?.getUsername(),
+  defaultHelp: async () => getDefaultDevopsCenterUsernameAlias(),
   required: true,
 });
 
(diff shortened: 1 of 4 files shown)


src/common/flags/flags.ts
```diff
@@ -47,7 +47,7 @@ export const requiredDoceOrgFlag = OclifFlags.custom({
   summary: messages.getMessage('flags.targetDoceOrg.summary'),
   parse: async (input: string | undefined) => getOrgOrThrow(input),
   default: async () => getOrgOrThrow(),
-  defaultHelp: async () => (await getOrgOrThrow())?.getUsername(),
+  defaultHelp: async () => getDefaultDevopsCenterUsernameAlias(),
```
EstebanRomero84, line 50: defaultHelp is just for the flag help and should not throw an exception when there is no default.

npiccolo (approved):
Looks good to me !! =D

PR #146, opened 2023-04-05
fix: add verbosity flags to resume command

EstebanRomero84:
### What does this PR do?
Adds verbosity flags to resume command.

[Insert GUS WI](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001NG4OJYA1/view)

### Functionality Before

**sf project deploy pipeline resume** does not have --concise and --verbose flags.

### Functionality After

**sf project deploy pipeline resume** has --concise and --verbose flags.

### DIFF
--- a/test/commands/project/deploy/pipeline/resume.test.ts
+++ b/test/commands/project/deploy/pipeline/resume.test.ts
@@ -205,7 +205,7 @@ describe('project deploy pipeline resume', () => {
         stubDisplayEndResults = sandbox.stub(ResumeCommandOutputService.prototype, 'displayEndResults');
       })
       .command(['project deploy pipeline resume', `-i=${mockAorId}`, '--verbose'])
-      .it('prints output for verbose flag correctly', () => {
+      .it('calls displayEndResults when deployment is completed', () => {
         // verify we printed the end results message
         expect(stubDisplayEndResults.called).to.equal(true);
       });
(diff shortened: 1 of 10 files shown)


test/commands/project/deploy/pipeline/resume.test.ts
```diff
@@ -177,6 +188,28 @@ describe('project deploy pipeline resume', () => {
...
+          sf_devops__Status__c: AsyncOperationStatus.InProgress,
+          sf_devops__Error_Details__c: 'mockErrorDetail',
+        };
+        sandbox.stub(AorSelector, 'selectAsyncOperationResultById').resolves(mockAorRecord);
+
+        monitorStub = sinon.stub(AsyncOpStreaming.prototype, 'monitor');
+        sandbox.stub(ResumeCommandOutputService.prototype, 'getStatus').returns(AsyncOperationStatus.Completed);
+        stubDisplayEndResults = sandbox.stub(ResumeCommandOutputService.prototype, 'displayEndResults');
+      })
+      .command(['project deploy pipeline resume', `-i=${mockAorId}`, '--verbose'])
```
JuanStenghele-sf, line 207: How about a test to check that stubDisplayEndResults is not called when the verbose flag is not used? And another one for the concise case (the output should be the same as if it was not used)?

JuanStenghele-sf (changes requested):
This looks really good! Just some test suggestions and we should also update the readme.md with the concise and verbose flags info.

test/commands/project/deploy/pipeline/resume.test.ts
EstebanRomero84: Good catch. I updated the tests and the READE file.
Thanks.

JuanStenghele-sf (approved):
Great job!

PR #162, opened 2023-04-27
juan/W-12651281

JuanStenghele-sf:
### What does this PR do?

Updates the CLI to the new promote API response.

[W-12651281](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001MgWKtYAN/view)

### Functionality Before

It handled a string API response.

### Functionality After

It now handles JSON responses.


src/common/base/abstractPromote.ts
```diff
@@ -207,7 +207,24 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
     };
-    return targetOrg.getConnection().request(req);
+    try {
+      const response: ApiPromoteResponse = await targetOrg.getConnection().request(req);
+      return response.jobId;
+    } catch (error) {
+      // The error message contains the API response
+      // We should parse that to get the error
+      const rawError = error as ApiError;
+      try {
```
JuanStenghele-sf: This second try is weird

src/common/types.ts
```diff
@@ -43,9 +43,14 @@ export enum TestLevel {
...
 export interface ApiError extends Error {
   errorCode: string;
-  name: string;
 }
 
+export interface ApiPromoteResponse {
+  jobId: string;
+}
+
+export type ApiQuickResponse = ApiPromoteResponse;
```
EstebanRomero84: Why do we need a specific type for quick? It is calling the same api and receiving the same response than the "regular" promote.

JuanStenghele-sf: Ok! We can use the same type. I made it like this to differentiate between the two operations, but we can change this.

EstebanRomero84 (approved):
This looks fine.
Thanks Juan.

PR #163, opened 2023-04-28
feat: add project deploy pipeline quick command

EstebanRomero84:
### What does this PR do?
Adds new command _project deploy pipeline quick_. Also moves the logic to monitor the async operation to a base class.

[W-12614670](https://gus.lightning.force.com/a07EE00001M9DKgYAN)

### Functionality Before

It is not posible to perform a quick deployment.

### Functionality After

It is posible to perform a quick deployment.


src/common/base/abstractAsyncOperation.ts
```diff
@@ -0,0 +1,109 @@
...
+  protected setOutputService(service: PromoteOutputService): void {
+    this.outputService = service;
+  }
+
+  /**
+   * Method to handle an in progress async operation. It stores the AOR in cache,
+   * then monitors the state of it and print messages if needed. Finally returns the
+   * final state of the operation.
+   */
+  protected async monitorOperation(isAsync: boolean, wait: Duration): Promise<AsyncOperationResultJson> {
```
EstebanRomero84, line 39: I am open to suggestions for a better method name.

npiccolo: inspectOperation ?

src/common/base/abstractAsyncOperation.ts
```diff
@@ -0,0 +1,109 @@
...
+  private async getOperationState(): Promise<AsyncOperationResultJson> {
+    const asyncJob: AsyncOperationResult = await fetchAsyncOperationResult(
+      this.targetOrg.getConnection(),
+      this.asyncOperationId
+    );
+    return {
+      jobId: this.asyncOperationId,
+      status: asyncJob.sf_devops__Status__c as string,
+      message: asyncJob.sf_devops__Message__c as string,
+      errorDetails: asyncJob.sf_devops__Error_Details__c as string,
```
npiccolo, line 91: If the Error_Details here is undefined or null you will get an error here I think.
There are some cases where Heroku throws an error and you don't get the Error_details.

src/common/base/abstractPromote.ts
```diff
@@ -73,77 +58,45 @@ export abstract class PromoteCommand<T extends typeof SfCommand> extends SfComma
...
-    this.outputService = new OutputServiceFactory().forDeployment(
-      this.flags,
-      (this.flags['devops-center-username'] as Org).getConnection()
+    this.setOutputService(
+      new OutputServiceFactory().forDeployment(
+        this.flags,
+        (this.flags['devops-center-username'] as Org).getConnection()
+      )
     );
+    this.targetOrg = this.flags['devops-center-username'] as Org;
```
npiccolo, line 61: Maybe move this before **this.setOutputService** and use **this.targetOrg.getConnection()** ??
So you don’t have to cast it twice

src/common/base/abstractQuick.ts
```diff
@@ -0,0 +1,171 @@
...
+
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('@salesforce/plugin-devops-center', 'project.deploy.pipeline.quick');
+
+export default abstract class QuickPromotionCommand<T extends typeof SfCommand> extends AsyncCommand {
+  public static baseFlags = {
+    async,
+    concise,
+    verbose,
+    wait,
```
npiccolo, line 37: I really like these are generic finally !!!

src/common/base/abstractQuick.ts
```diff
@@ -0,0 +1,171 @@
...
+      new OutputServiceFactory().forQuickDeployment(
+        this.flags,
+        (this.flags['devops-center-username'] as Org).getConnection(),
+        this.targetStage.sf_devops__Branch__r.sf_devops__Name__c
+      )
+    );
+
+    this.setAsyncOperationId(
+      await this.requestPromotion(
+        this.targetStage?.sf_devops__Pipeline__r.sf_devops__Project__c,
```
npiccolo: This can be null/undefined ???

src/common/base/abstractAsyncOperation.ts
EstebanRomero84: That would be similar to a successful promotion, where  _sf_devops__Error_Details__c_ is null right? I tested it and it doesn't throw.

src/common/base/abstractPromote.ts
EstebanRomero84: Good idea. I will change it.

src/common/base/abstractQuick.ts
EstebanRomero84: Nop.

npiccolo: Why we have the **?** ?

src/common/base/abstractAsyncOperation.ts
npiccolo: Yes, you will have a status = ERROR and no Error_Details at all.
I had that issue and it doesn’t print anything on the screen about the error because it was not handled on aorOutputService printAorStatus.
On my future PR I have a fix for that, we are going to print the error even if we don’t have error details.

src/common/base/abstractQuick.ts
EstebanRomero84: It's not needed, I will remove it. Thanks.

src/commands/project/deploy/pipeline/quick.ts
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Computes the source stage Id for quick promotions. Quick promotions are allowed only
+   * in versioned promotions so source stage is previous stage.
+   *
+   * @returns: string. It is the source stage Id.
+   */
+  protected getSourceStageId(): string {
+    // for versioned promotions we always have a previous stage
+    const prevStages = this.getTargetStage().sf_devops__Pipeline_Stages__r as PreviousPipelineStages;
+    return prevStages.records[0].Id;
```
molson-sfdc, line 38: I don't see anywhere we enforce getTargetStage() returns a versioned stage.  So they could specify the first stage and get a list index error here...right?  Or did I miss something.

src/common/base/abstractQuick.ts
molson-sfdc: :same:

molson-sfdc (commented):
Just one observation.  The rest looks really good.  I like the new abstraction layer, makes sense.

src/commands/project/deploy/pipeline/quick.ts
EstebanRomero84: The users don't manually provide the target stage, they provide the AOR Id. So we get the target stage through the relation AOR -> Deployment Result  -> CBI -> Environment -> Stage.
We do validate that the AOR correspond to a validated deployment and since we allow validation deployment only for versioned promotions we are implicitly validating that the target stage has a previous stage.
If in the future we allow validation deployments for ad hoc promotions we will have to update this method.

npiccolo (approved):
Looks good to me !!
Thanks Esteban !

PR #164, opened 2023-04-28
Npiccolo/validate deploy

npiccolo:
### What does this PR do?

Adds validate-only deployment command + functionality 

@[W-12468380](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001KPXmfYAH/view)@

### Functionality Before

No command

### Functionality After

Validate-only Deployment


src/commands/project/deploy/pipeline/validate.ts
```diff
@@ -0,0 +1,35 @@
...
+import { PromoteOptions, PromotePipelineResult } from '../../../../common';
+import DeployPipeline from './start';
+
+Messages.importMessagesDirectory(__dirname);
+const messages = Messages.loadMessages('@salesforce/plugin-devops-center', 'project.deploy.pipeline.validate');
+
+/**
+ * Contains the logic to execute the sf project deploy pipeline start command.
+ */
+export default class validateDeployPipeline extends DeployPipeline {
```
molson-sfdc:
```suggestion
export default class ValidateDeployPipeline extends DeployPipeline {
```

Let's stay consistent

src/common/outputService/deploySummaryBuilder.ts
```diff
@@ -269,3 +273,73 @@ class SoupDeploySummaryOutputService<T extends OutputFlags> extends VersionedOrS
     this.workItems = workItems;
   }
 }
+
+/**
+ * Service class used to print the operation summary of an validate-deploy
+ *
+ */
+class ValidateDeploySummaryOutputService<T extends OutputFlags> extends DeploySummaryOutputService<T> {
```
molson-sfdc: Just a thought...maybe add "Versioned" into this name....someday we might support AdHoc, which will be different

src/common/selectors/deployComponentsSelector.ts
```diff
@@ -16,7 +16,8 @@ export async function selectDeployComponentsByAsyncOpId(
 ): Promise<DeployComponent[]> {
   const queryStr = `SELECT sf_devops__Source_Component__c, sf_devops__Operation__c, sf_devops__File_Path__c
                     FROM sf_devops__Deploy_Component__c  
-                    WHERE sf_devops__Deployment_Result__r.sf_devops__Status__c = '${asyncOpId}'`;
+                    WHERE sf_devops__Deployment_Result__r.sf_devops__Status__c = '${asyncOpId}'
+                    OR sf_devops__Deployment_Result__r.sf_devops__Check_Deploy_Status__c = '${asyncOpId}'`;
```
molson-sfdc: It's early, and I have not had all my coffee yet so I cannot think of a concrete case, but this seems error prone.  Is there a case where we would get the wrong set of Deploy_Components because of the OR?

Do we just want to create a second selector to be more specific?  selectDeployComponentsForCheckDeployByAsynchOpId?

molson-sfdc (commented):
Looks good.  Just a few minor comments

src/commands/project/deploy/pipeline/validate.ts
```diff
@@ -0,0 +1,35 @@
...
+  public static readonly examples = messages.getMessages('examples');
+  public static readonly state = 'beta';
+  private readonly isCheckDeploy = true;
+
+  public async run(): Promise<PromotePipelineResult> {
+    return this.executePromotion();
+  }
+
+  /**
+   * Returns the promote option necessary to performs an undeployed only promotion
```
EstebanRomero84:
```suggestion
   * Returns the promote option necessary to performs an check deploy promotion
```

src/common/selectors/deployComponentsSelector.ts
npiccolo: Yes !!

src/common/outputService/deploySummaryBuilder.ts
```diff
@@ -269,3 +273,73 @@ class SoupDeploySummaryOutputService<T extends OutputFlags> extends VersionedOrS
...
+  const pipelineInfo: PipelineInfo = {
+    stageName: envQueryResp.sf_devops__Pipeline_Stages__r.records[0].Name,
+    environmentName: envQueryResp.Name,
+  };
+
+  return pipelineInfo;
+}
+
+// This type will be defined here as it is sepecific from this function
+export type PipelineInfo = {
```
EstebanRomero84:
```suggestion
export type PipelineStageInfo = {
```

src/commands/project/deploy/pipeline/validate.ts
```diff
@@ -0,0 +1,35 @@
...
+   */
+  protected getPromoteOptions(): Partial<PromoteOptions> {
+    return { checkDeploy: this.isCheckDeploy };
```
EstebanRomero84: As mentioned in DM, this needs to include `undeployedOnly: true`

molson-sfdc: It should get all of the options from the super class...right?

npiccolo: Yes !!!!
Esteban suggested something like this:

_return {…super.getPromoteOptions(), checkDeploy: this.isCheckDeploy}_

EstebanRomero84 (approved):
This looks fine to me.

EstebanRomero84 (approved):
Looking good.
Thanks Nico.

PR #179, opened 2023-05-18
Npiccolo/stall

npiccolo:
### What does this PR do?
Handles fast transition to concluded state for an AOR during any async operation running on the CLI.

@[W-13065560](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001Pp0vJYAR/view)@

### Functionality Before

Not handled

### Functionality After

If transitioned fast, we display the status of the AOR.

### DIFF
--- a/src/common/base/abstractAsyncOperation.ts
+++ b/src/common/base/abstractAsyncOperation.ts
@@ -98,12 +98,23 @@ export abstract class AsyncCommand extends SfCommand<AsyncOperationResultJson> {
    * @param wait
    */
   private async startMonitoring(wait: Duration): Promise<void> {
-    const streamer: DoceMonitor = getAsyncOperationStreamer(
-      this.targetOrg,
-      wait,
-      this.asyncOperationId,
-      this.outputService
+    const asyncJob: AsyncOperationResult = await fetchAsyncOperationResult(
+      this.targetOrg.getConnection(),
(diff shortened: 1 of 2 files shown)


src/common/base/abstractAsyncOperation.ts
```diff
@@ -98,12 +98,23 @@ export abstract class AsyncCommand extends SfCommand<AsyncOperationResultJson> {
...
-      wait,
-      this.asyncOperationId,
-      this.outputService
+    const asyncJob: AsyncOperationResult = await fetchAsyncOperationResult(
+      this.targetOrg.getConnection(),
+      this.asyncOperationId
     );
-    await streamer.monitor();
+    if (
+      asyncJob.sf_devops__Status__c === undefined ||
```
EstebanRomero84, line 106: In which case sf_devops__Status__c could be undefined? The field is a required picklist so should always have value right?

EstebanRomero84 (approved):
Looks good, just one question.

src/common/base/abstractAsyncOperation.ts
npiccolo: I was respecting the aorOutputService logic we currently have(I have updated it and now is not explicitly checked **== undefined**, but it was like that before)
What you say makes sense.
We can remove this undefined check maybe

PR #222, opened 2023-08-28
chore: remove caret on oclif/core dev dependency

EstebanRomero84:
### What does this PR do?
Remove caret from `@oclif/core` dev dependency so it can be used in CodeBuilder.

[W-14026594](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001ZXMS4YAP/view)

### Functionality Before

Plugin commands fail with error `TypeError: this.config.getPluginsList is not a function` when executed in CodeBuilder.

### Functionality After

Commands run as expected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "author": "Salesforce",
   "bugs": "https://github.com/forcedotcom/cli/issues",
   "dependencies": {
-    "@oclif/core": "^2.11.5",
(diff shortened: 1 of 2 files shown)


shetzel (changes requested):
You need to run `yarn install` after making this change so that `yarn.lock` is updated.

shetzel (approved):

PR #232, opened 2023-09-01
Fix for quick deployment

EstebanRomero84:
### What does this PR do?
Adds validations for two error cases:

- Validate that the deployment result for a quick deploy has CBI associated.
- Validate that we don't print undefined when monitoring the AOR progress.

[@W-14002948@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001Z9PyCYAV/view)
[@W-14002980@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001Z9ifMYAR/view)

### DIFF
--- a/src/common/outputService/aorOutputService.ts
+++ b/src/common/outputService/aorOutputService.ts
@@ -76,8 +76,7 @@ export abstract class AbstractAorOutputService<T extends AorOutputFlags>
       } else {
         console.log(red(aor.sf_devops__Message__c));
       }
-    } else {
-      // aor.sf_devops__Status__c == undefined || AsyncOperationStatus.InProgress
+    } else if (aor.sf_devops__Message__c) {
       console.log(aor.sf_devops__Message__c);
     }
 
--- a/src/common/selectors/deploymentResultsSelector.ts
+++ b/src/common/selectors/deploymentResultsSelector.ts
@@ -62,9 +62,11 @@ export async function selectOneDeploymentResultWithChangeBundleInstallsByAsyncJo
   asyncJobId: string
 ): Promise<CheckDeploymentResultWithChangeBundleInstalls | null> {
   const queryStr = `SELECT sf_devops__Check_Deploy__c, sf_devops__Deployment_Id__c, sf_devops__Check_Deploy_Status__r.sf_devops__Status__c,
-                    (SELECT sf_devops__Environment__c FROM sf_devops__Change_Bundle_Installs__r)
+                      (SELECT sf_devops__Environment__c FROM sf_devops__Change_Bundle_Installs__r)
                     FROM sf_devops__Deployment_Result__c  
-                    WHERE sf_devops__Check_Deploy_Status__c = '${asyncJobId}' AND sf_devops__Check_Deploy__c = TRUE`;
+                    WHERE sf_devops__Check_Deploy_Status__c = '${asyncJobId}'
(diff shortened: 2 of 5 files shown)


src/common/outputService/aorOutputService.ts
```diff
@@ -76,8 +76,7 @@ export abstract class AbstractAorOutputService<T extends AorOutputFlags>
       } else {
         console.log(red(aor.sf_devops__Message__c));
       }
-    } else {
-      // aor.sf_devops__Status__c == undefined || AsyncOperationStatus.InProgress
+    } else if (aor.sf_devops__Message__c) {
```
franciscoperezsammartino, line 79: I see line 77 and 68 that also print sf_devops__Message__C without checking it is null. Is that okey?

src/common/selectors/deploymentResultsSelector.ts
```diff
@@ -62,9 +62,11 @@ export async function selectOneDeploymentResultWithChangeBundleInstallsByAsyncJo
   asyncJobId: string
 ): Promise<CheckDeploymentResultWithChangeBundleInstalls | null> {
   const queryStr = `SELECT sf_devops__Check_Deploy__c, sf_devops__Deployment_Id__c, sf_devops__Check_Deploy_Status__r.sf_devops__Status__c,
-                    (SELECT sf_devops__Environment__c FROM sf_devops__Change_Bundle_Installs__r)
+                      (SELECT sf_devops__Environment__c FROM sf_devops__Change_Bundle_Installs__r)
                     FROM sf_devops__Deployment_Result__c  
-                    WHERE sf_devops__Check_Deploy_Status__c = '${asyncJobId}' AND sf_devops__Check_Deploy__c = TRUE`;
+                    WHERE sf_devops__Check_Deploy_Status__c = '${asyncJobId}'
+                      AND sf_devops__Check_Deploy__c = TRUE
+                      AND Id IN (SELECT sf_devops__Deployment_Result__c FROM sf_devops__Change_Bundle_Install__c)`;
```
franciscoperezsammartino, line 69: Just verifying a CBI exists is enough? or should we verify the status is completed.

Maybe that is handled elsewhere, and just needed to be sure CBI was present

franciscoperezsammartino (commented):
Looks good, two questions

src/common/selectors/deploymentResultsSelector.ts
EstebanRomero84: Yes, the status is validated in `abstractQuick.ts`. We don't filter the query by that so we get the current status and show a proper error message if needed.
For the case of missing CBIs we should show the same error message as if the deployment result were null so that's way added the filter to the query.

src/common/outputService/aorOutputService.ts
EstebanRomero84: When we update the Status field to Complete or to Error we also update the Message field, in the same transaction. So I think for those to cases the Message field will always have a value.
When the Status is In Progress it is posible that Message is null because we can be updating other fields and not Message.

Do you think there could be a scenario where the Status is Complete|Error and Message is null? If so I can add the check there too.

franciscoperezsammartino: Yes, completed for sure has. And error we have logic in place to either log the message, if not the exception message, if not the exception so that should cover it.

If we see something we can revisit this.

franciscoperezsammartino (approved):

PR #292, opened 2023-11-28
Quick deployments from bundling stage

EstebanRomero84:
### What does this PR do?
Update quickPromotion to not send testLevel. This will allow to perform quick promotions from the bundling stage.

@W-14338568@

### Functionality Before
We send testLevel on quick promotion requests. This makes the transaction to fail due to new validation.

### Functionality After

We don't send testLevel on quick promotion requests.


src/common/base/abstractQuick.ts
```diff
@@ -149,7 +149,6 @@ export default abstract class QuickPromotionCommand<T extends typeof SfCommand>
    */
   private buildPromoteOptions(deployId: string): void {
     this.deployOptions = {
-      testLevel: 'Default',
```
EstebanRomero84, line 152: This is the main change: we don't need to send testLevel on a quick promotion request.

src/configMeta.ts
```diff
@@ -5,7 +5,7 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import { spawnSync } from 'child_process';
+import { spawnSync } from 'node:child_process';
```
EstebanRomero84, line 8: Needed to solve eslint error after dev dependencies update.

molson-sfdc (approved):

src/common/base/abstractQuick.ts
npiccolo: The testLevel is optional right ?

franciscoperezsammartino: Nice find

EstebanRomero84: We don't need to specify a test level on a quick deploy since it only deploy metadata, it does't run the tests. They were run in the previous validate-only deployment.

PR #523, opened 2026-09-22
@W-24217700 - work-item update/list fixes + stage branch/environment orphan guards

ad-shreya:
@W-24217700@
### What does this PR do?
This PR bundles several DevOps Center CLI fixes and enhancements around work items, the work-item list output, and adding environments to pipeline stages.

- Enforce work-item status transitions. A work item can now move to IN_PROGRESS only from NEW, and to READY_TO_PROMOTE only from IN_REVIEW. Invalid transitions are rejected with a clear, descriptive error before any update is sent.
- Fix work-item update --description / --subject. These are plain WorkItem sObject fields — the connect work-item endpoint only accepts status — so they are now written via the sObject API, fixing the Unrecognized field "description" error. Status changes still go through the connect endpoint so platform-side status handling runs.
- Show work item ID in work-item list. Added an ID column to the table so the record ID is visible alongside the name (the ID was already present in the query and --json output).
- Fix misleading target branch for terminal-stage work items. The list previously fell back to the pipeline's first-stage branch whenever the next-stage lookup was empty, so a work item already on the last stage (e.g. a CLOSED item) wrongly displayed the first stage's branch. Now the first-stage fallback applies only when a work item hasn't entered the pipeline; a work item on a stage with no next stage correctly shows a blank target branch.
- Prevent orphaning environments in stage environment add. Adding an environment to a stage that already has one previously created a new DevopsEnvironment and re-pointed the stage's lookup, orphaning the old record. The command now blocks by default with an actionable error, and a new --force flag removes the existing environment before adding the new one (aborting if that removal fails).
- Snapshot update. Regenerated command-snapshot.json for the new stage environment add --force flag.
#<Insert GitHub Issue>, @<Insert GUS WI>@

### Functionality Before

<insert gif and/or summary>

### Functionality After

<insert gif and/or summary>


src/utils/updateWorkItem.ts
```diff
@@ -81,32 +121,48 @@ export async function resolveProjectIdForWorkItem(connection: Connection, workIt
...
  */
 export async function updateWorkItem(params: UpdateWorkItemParams): Promise<UpdateWorkItemResult> {
   const { connection, workItemId, projectId, status, subject, description } = params;
 
-  const path = `/services/data/v${connection.getApiVersion()}/connect/devops/projects/${projectId}/workitem/${workItemId}`;
-  const payload: Record<string, string> = {};
-  if (status !== undefined) payload.status = toApiStatus(status);
-  if (subject !== undefined) payload.subject = subject;
-  if (description !== undefined) payload.description = description;
+  if (subject !== undefined || description !== undefined) {
```
anuragbhoumick: [WARNING] §00-common/Error Handling — Partial update when both fields and status are changed

**Context:** When the caller passes `--subject` (or `--description`) together with `--status`, the sObject write for subject/description runs first (lines 134–146). The status transition check (`assertStatusTransitionAllowed`) runs later at line 151.

**Issue:** If the sObject write succeeds but the transition check then throws (work item in a disallowed state), subject/description are already persisted while the function throws an error. The caller sees a failure, but the subject change silently stuck.

**Suggestion:** Move `assertStatusTransitionAllowed` (and `toApiStatus`) to run _before_ the sObject write — it's a read-only guard, so reordering is safe:
```ts
let apiStatus: string | undefined;
if (status !== undefined) {
  apiStatus = toApiStatus(status);
  await assertStatusTransitionAllowed(connection, workItemId, apiStatus);
}
// now safe to write fields
if (subject !== undefined || description !== undefined) { … }
// then the Connect API call
if (apiStatus) { … }
```

src/commands/devops/promote.ts
```diff
@@ -211,6 +224,27 @@ export default class DevopsPromote extends SfCommand<PromoteResult> {
...
+   * Blocks promotion if another promotion to the same target stage is already in flight, to
+   * avoid creating a duplicate. Scoped to the target stage (and therefore its pipeline), so
+   * promotions in other stages or pipelines are unaffected. Best-effort: if the check can't
+   * run, the promotion proceeds as before.
+   */
+  private async assertNoInFlightPromotion(connection: Connection, targetStageId: string): Promise<void> {
+    let inFlight: InFlightPromotion[];
+    try {
+      inFlight = await findInFlightPromotions(connection, targetStageId);
+    } catch {
```
anuragbhoumick: [WARNING] §00-common/Error Handling — Silent exception in catch block

**Context:** `assertNoInFlightPromotion` wraps `findInFlightPromotions` in a try/catch so a guard failure doesn't break the promote flow.

**Issue:** The catch block returns silently with no logging. Per §00-common: "Never swallow exceptions silently — at minimum log a warning." A network error, permissions issue, or SOQL failure here would be invisible, making it hard to diagnose why a duplicate promotion wasn't blocked.

**Suggestion:** Add a warning log so operators can see when the guard was skipped:
```ts
} catch (err) {
  this.warn(`In-flight promotion check skipped: ${err instanceof Error ? err.message : String(err)}`);
  return;
}
```

src/utils/promoteStage.ts
```diff
@@ -81,3 +82,47 @@ export async function promoteStage(params: PromoteStageParams): Promise<PromoteS
...
+ * block a duplicate promotion to the same stage while one is already running, without
+ * affecting promotions in other stages or pipelines.
+ */
+export async function findInFlightPromotions(
+  connection: Connection,
+  targetStageId: string
+): Promise<InFlightPromotion[]> {
+  validateSalesforceId(targetStageId, 'target stage');
+  const statusList = IN_FLIGHT_PROMOTION_STATUSES.join("', '");
+  const result = await connection.query<PipelineStagePromotionRecord>(
```
anuragbhoumick: [NOTE] §00-common/Code Quality — SOQL query without LIMIT clause

**Context:** `findInFlightPromotions` queries `DevopsPipelnStgProm` filtered by stage + status with no LIMIT.

**Issue:** While in practice there should be 0–1 in-flight promotions per stage, an unbounded query in a governor-limited context could be expensive with unexpected data.

**Suggestion:** Add `LIMIT 10` as a safety net, or `LIMIT 1` and return a boolean since the caller only checks `length > 0`.

anuragbhoumick (commented):
**Review Summary**: 2 warnings, 1 note across 31 files (1265+/51−). Well-structured PR — the status transition enforcement, orphan-prevention guards, and in-flight promotion check are solid additions with clean test coverage. Two ordering/error-handling concerns worth addressing.

